Issue #371, opened 2026-04-28
TestSessionRegistryRoundtrip is a stub — session SQL never verified in CI


petersimmons1972:
## Problem
`internal/db/postgres_session_test.go` contains `TestSessionRegistryRoundtrip` which acquires `TEST_DATABASE_URL` and then does nothing (`_ = dsn`, `// TODO` comment). The four session persistence methods (`RegisterSession`, `UnregisterSession`, `ListActiveSessions`, `TouchSession`) have zero integration coverage. A typo in the SQL or a type mismatch would pass all CI gates.

## Fix
Implement the roundtrip test against a real Postgres instance. The existing `testutil.DSN(t)` helper handles the skip-if-unset behavior. Follow the pattern in `internal/db/postgres_document_test.go`.

Reported by New Maintainer, Skeptical Staff Engineer, Operator SRE, Heavy CLI User (4/6 personas).

petersimmons1972:
Fixed in #373 (merged to main at 34cdc8e).

Issue #370, opened 2026-04-28
GlobalReembedder: per-chunk embed failures not counted; efSearch uncapped at 1000


petersimmons1972:
### 1. Per-chunk embed failures swallowed
In `global_worker.go`, embed errors are logged at `Warn` and counted as zero (non-fatal). There is no Prometheus counter increment per failed chunk. `metrics.WorkerErrors.WithLabelValues("global_reembed")` is only incremented on outer `runBatch` errors, not per-chunk failures. A batch that fails to embed 90 of 100 chunks looks identical to a successful batch in metrics.

Fix: add `metrics.WorkerErrors.WithLabelValues("global_reembed").Inc()` inside the embed-failure branch of the errgroup goroutine.

### 2. hnsw.ef_search uncapped
`VectorSearch` sets `SET LOCAL hnsw.ef_search = limit * 2`. With `ENGRAM_RECALL_MAX_TOP_K=500` and `topK*3` passed to `VectorSearch`, the maximum ef_search is 3000. pgvector's documented maximum is 1000. Values above that are silently clamped or cause warnings depending on version.

Fix: add `if efSearch > 1000 { efSearch = 1000 }` before the SET LOCAL.

Reported by Operator SRE, Heavy CLI User (per-chunk failures), Skeptical Staff Engineer (ef_search cap).

petersimmons1972:
Fixed in #373 (merged to main at 34cdc8e).

Issue #369, opened 2026-04-28
Auto-tag preference: false positives on common engineering phrases


petersimmons1972:
## Problem
`IsPreferenceContent` in `internal/search/query_signal.go` matches on tokens including "like" and "want". Common engineering phrases trigger false auto-tagging:

- "I would like to document the architecture decision" → tagged `preference` instead of `decision`
- "We want to ensure the cache is invalidated correctly" → tagged `preference` instead of `context`

Memories incorrectly tagged as `preference` receive a 1.2× boost on preference queries and are demoted on decision/context queries — the inverse of what the user wants.

## Fix options
- Narrow the signal list to higher-specificity tokens ("prefer", "favorite", "favourite") — remove "like" and "want"
- Or require first-person possessive context ("I like", "I want", "my favorite") rather than bare token matching

Reported by Skeptical Staff Engineer, Docs-First Newcomer (2/6 personas).

petersimmons1972:
Fixed in #373 (merged to main at 34cdc8e).

Issue #368, opened 2026-04-28
Three new env vars undocumented: ENGRAM_REEMBED_BATCH_SIZE, ENGRAM_REEMBED_INTERVAL, ENGRAM_RECALL_MAX_TOP_K


petersimmons1972:
## Problem
PR #366 introduced three new runtime-configurable env vars that are absent from `.env.example`, `docs/getting-started.md` (configuration table), and `docs/operations.md`:

- `ENGRAM_REEMBED_BATCH_SIZE` (default: 100) — global reembedder batch size
- `ENGRAM_REEMBED_INTERVAL` (default: 10s) — global reembedder poll interval  
- `ENGRAM_RECALL_MAX_TOP_K` (default: 500) — maximum allowed top_k for memory_recall

## Fix
Add all three to `.env.example` with comments and to the configuration reference table in `docs/getting-started.md`.

Reported by Docs-First Newcomer, Heavy CLI User (2/6 personas).

petersimmons1972:
Fixed in #373 (merged to main at 34cdc8e).

PR #372, opened 2026-04-28
fix(metrics): restore ChunksPendingReembed gauge in GlobalReembedder

petersimmons1972:
## Summary

- Restores `metrics.ChunksPendingReembed` gauge update that was dropped when `GlobalReembedder` replaced per-project `reembed.Worker` in PR #366
- The old `Worker` updated this gauge at `worker.go:144` before each batch; `GlobalReembedder` never carried that responsibility over, leaving the gauge permanently stale

## Root cause

Found by the Operator / SRE persona in the 6-person QA sweep run after #366 merged. Classified as blocker — dashboards and `engram_chunks_pending_reembed` alerts are blind to embedding backlog after this change.

## Fix

Query `SELECT COUNT(*) FROM chunks WHERE embedding IS NULL` with a 5-second deadline before each batch and call `metrics.ChunksPendingReembed.Set(float64(pending))`. One-line behaviour change; identical to what `worker.go` does.

## Test plan

- [ ] `go test ./internal/reembed/... -count=1 -race` — passes
- [ ] `go test ./... -count=1 -race` — 37/37 packages pass

## Related

Remaining non-blocking findings from the QA sweep tracked in: #367 #368 #369 #370 #371

### DIFF
--- a/cmd/engram/main.go
+++ b/cmd/engram/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/petersimmons1972/engram/internal/entity"
 	internalmcp "github.com/petersimmons1972/engram/internal/mcp"
 	"github.com/petersimmons1972/engram/internal/netutil"
+	"github.com/petersimmons1972/engram/internal/reembed"
 	"github.com/petersimmons1972/engram/internal/search"
 	"github.com/petersimmons1972/engram/internal/summarize"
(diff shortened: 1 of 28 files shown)


petersimmons1972:
Superseded by #374 — clean cherry-pick onto current main. PR #372 diverged before the PR #366 squash merge, making direct merge impossible without re-applying already-landed changes.

PR #432, opened 2026-05-04
fix(#430): support cross-project federated edges and conflict enrichment

### LINKED ISSUE #430: Cross-project relationship support: StoreRelationship + GetMemory should permit federated edges
Spun off from #429 as a focused follow-up. The other three tests in #429 are now fixed.

## What's needed
`TestFederatedRecall_IncludeConflicts_NotSilentlyDropped` (currently `t.Skip`'d at `internal/mcp/federated_conflicts_test.go:32`) requires two product-level changes:

1. **`PostgresBackend.StoreRelationship`** — drop the project filter from the source/target existence checks (`internal/db/postgres_relationship.go` lines 26–43). Memory IDs are global UUIDs, so the project filter rejects valid cross-project edges (memA in proj1 → memB in proj2) without preventing any real ID-collision risk. The edge's own `rel.Project` field still records ownership.

2. **`PostgresBackend.GetMemory` (or a new `GetMemoryByID`)** — `EnrichWithConflicts` (`internal/mcp/conflicts.go:101`) needs to fetch the other endpoint of a contradicts edge regardless of project. Two options:
   - **A:** drop the project filter on `GetMemory` directly. Smallest diff; affects every caller. Justified because IDs are UUIDs.
   - **B:** add `GetMemoryByID(ctx, id)` that does not filter by project, and use it only in `EnrichWithConflicts`. Larger diff, narrower blast radius.

## Tradeoff
- **Pro:** federated recall surfaces cross-project contradictions, which is the whole point of `include_conflicts=true` on a multi-project query.
- **Con:** any caller that mistakenly passes an ID from another project will silently get that other project's row instead of `nil`. This is a defense-in-depth change.
- **Neutral:** tenancy is still enforced at `GetRelationships` (project-scoped) and at every project-scoped query (`SELECT ... WHERE project=`). UUIDs are not guessable, so dropping the filter on by-ID lookups doesn't enable new attacks.

## Recommendation (lean)
Option A. One-line change in `PostgresBackend.GetMemory` + drop the existence-check project filter in `StoreRelationship`. Document both. Un-skip the test.

## Acceptance
- `TestFederatedRecall_IncludeConflicts_NotSilentlyDropped` passes without t.Skip.
- All existing tests continue to pass.
- TDD red-then-green commit.

petersimmons1972:
## Summary
- Add `GetMemoryByID(ctx, id)` to `Backend` interface — fetches memory by global UUID without project filter
- Drop project filter from `StoreRelationship` source/target existence checks (memory IDs are global UUIDs)
- `EnrichWithConflicts` now uses `GetMemoryByID` so it can surface contradicting memories from other projects
- Add `GetMemoryByID` stubs to all test doubles (no production logic change in tests)

Closes #430.

Implements **Option B** from the issue (narrower blast radius than dropping the filter on `GetMemory` directly). The acknowledged tradeoff: a caller passing a wrong-project ID into `EnrichWithConflicts` would get that other project's row. This is acceptable because IDs are global UUIDs and the call site is constrained.

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./...` — all 38 packages pass
- [ ] `TestFederatedRecall_IncludeConflicts_NotSilentlyDropped` passes against a real database (requires `TEST_DATABASE_URL`)


petersimmons1972:
## Adversarial review status (per repo CLAUDE.md AI-Generated PR Policy)

- ✅ **Zero-context structural review** — completed self-review against the issue text. Verdict: PASS-WITH-NITS. Single nit: `GetMemoryByID` uses `SELECT *` (fragile to schema evolution but uses same `rowToMemory` decoder as `GetMemory`). Non-blocking.
- ⏳ **Rickover correctness audit** — pending. Sub-agent dispatch was rate-limited (resets May 6 06:00 ET).
- ⏳ **Spruance coverage audit** — pending. Same rate-limit blocker.

**Branch protection** also requires 1 approving human review and a status check named `test` (current actual check is `Test & Coverage` — possible naming drift).

Holding for completion of Rickover + Spruance before merge per repo policy. `severity/blocker` finding gate is not yet satisfied.

petersimmons1972:
## Adversarial review complete

- ✅ **Rickover (correctness)**: CERTIFIED — 0 blockers. All gates passed: SQL injection, parameter mismatch, error handling, nil-handling, soft-delete parity, integrity-check parity, StoreRelationship tx-safety. Single non-blocking observation (helper-extraction) noted for future cycle.
- ✅ **Spruance (coverage)**: Initial pass found 2 blockers: missing TestStoreRelationship_CrossProject and missing direct GetMemoryByID tests. Both **remediated in commit 73d7868** — see internal/db/postgres_memory_test.go and the addition to internal/db/postgres_relationship_test.go.
- ✅ **Zero-context structural review**: PASS-WITH-NITS — single nit (`SELECT *` in GetMemoryByID, non-blocking).

Coverage gates met. Ready for human approval per branch protection.

Issue #765, opened 2026-05-20
nice-to-have: memory_correct does not recalculate valid_from when date: tags change


petersimmons1972:
## Summary

`handleMemoryCorrect` / `UpdateMemory` does not recalculate `valid_from` when tags are updated. If a caller adds or removes a `date:` tag via `memory_correct`, the stored `valid_from` value in the DB is not updated.

## Impact

Low: callers using `memory_correct` to add a temporal anchor after the fact will see stale `valid_from`. `temporalAnchorHours` will continue using the old value (or NULL → LastAccessed fallback). Not silent at store time — only affects post-hoc corrections.

## Proposed Fix

In `UpdateMemory` (or its caller `engine.Correct`), if tags are updated, re-run `parseDateTag(newTags)` and add `valid_from = $N` to the UPDATE statement.

Alternatively, document explicitly that `valid_from` is immutable after initial store.

## Affected Code

- `internal/db/postgres_memory.go` lines 229–237 (UpdateMemory UPDATE statements, SHA c409c22)
- `internal/mcp/tools_store.go` `handleMemoryCorrect` (SHA c409c22)

Issue #764, opened 2026-05-20
nice-to-have: memory_store_batch silently drops immutable=true per-item flag


petersimmons1972:
## Summary

`handleMemoryStoreBatch` does not read the per-item `immutable` field from each batch item. The `types.Memory` literal at `internal/mcp/tools_store.go` lines 300–310 always constructs memories with `Immutable: false` (zero value), regardless of what the caller sends.

The tool docstring says "Each item supports the same optional fields as memory_store", and `handleMemoryStore` line 89 reads `getBool(args, "immutable", false)` — so callers reasonably expect per-item `immutable` to work.

## Impact

Low: immutable flag is rarely used per-item in batch calls. However, it means callers cannot atomically batch-store a mix of immutable and mutable memories. Any item with `"immutable": true` silently degrades to mutable.

## Proposed Fix

In the per-item memory construction loop, add:
```go
Immutable: getBool(mmap, "immutable", false),
```
matching `handleMemoryStore` line 89.

## Affected Code

- `internal/mcp/tools_store.go` lines 300–310 (SHA c409c22)

Issue #763, opened 2026-05-20
bug: memory_store_document drops valid_from and episode_id — temporal anchor and episode linkage silently lost


petersimmons1972:
## Summary

`handleMemoryStoreDocument` in `internal/mcp/tools_store.go` (lines 155–216) constructs a `types.Memory` struct (lines 185–192) that is missing two fields that `handleMemoryStore` correctly populates:

1. **`ValidFrom`** — never set; `parseDateTag(docTags)` is not called. Same class of bug as #747 and #762.
2. **`EpisodeID`** — `episodeIDFromContextOrArgs(ctx, args)` is never called, so documents stored during a named episode are not linked to that episode.

## Impact

**`ValidFrom`:** Documents tagged with `date:` tags (or ingested via tools that prepend date tags) will have `nil` ValidFrom. `temporalAnchorHours` falls back to `LastAccessed` (ingest time), silently discarding the intended temporal anchor and degrading recency scoring for all document-mode memories.

**`EpisodeID`:** Any document stored while an episode is active (e.g. during a named work session) will NOT appear in `memory_episode_recall` results. The episode–memory linkage is silently dropped.

## Affected Code

- `internal/mcp/tools_store.go` lines 185–192 (SHA c409c22)
- `handleMemoryStore` lines 82–94 shows the correct pattern with both fields populated

## Proposed Fix

```go
m := &types.Memory{
    // ... existing fields ...
    EpisodeID: episodeIDFromContextOrArgs(ctx, args),  // ADD
    ValidFrom: parseDateTag(docTags),                   // ADD
}
```

petersimmons1972:
Fixed in commit 3178f08 (PR #778). The fix wires `EpisodeID: episodeIDFromContextOrArgs(ctx, args)` and `ValidFrom: parseDateTag(docTags)` into `handleMemoryStoreDocument`'s `types.Memory` literal, matching the `handleMemoryStore` pattern. Regression test `TestMemoryStoreDocument_PersistsValidFromAndEpisode` added in `internal/mcp/tools_store_test.go` — passes green with `go test -race ./...`.

Issue #762, opened 2026-05-20
bug: memory_store_batch drops valid_from (date: tags not parsed for batch items)


petersimmons1972:
## Summary

`handleMemoryStoreBatch` in `internal/mcp/tools_store.go` never calls `parseDateTag` on per-item tags, so `Memory.ValidFrom` is always `nil` for every batch-stored memory — even when the caller passes a `date:` tag.

The single-item path (`handleMemoryStore`, line 92) correctly calls `parseDateTag(tags)` and assigns `ValidFrom`. The batch path (lines 300–310) omits this entirely.

## Impact

`ValidFrom` is the primary input to `temporalAnchorHours` in `internal/search/engine.go:1030-1034`. When `ValidFrom` is nil, the scorer falls back to `LastAccessed` (set to NOW() at ingest). For historical/dated memories stored in bulk (e.g. LongMemEval haystack ingestion), this silently zeroes out the temporal anchor, causing all batch-ingested dated memories to appear equally recent regardless of their `date:` tag.

This is the same class of silent regression as #747.

## Repro

```go
// Batch item with date: tag
{"content": "...", "tags": ["date:2023-01-01", "foo"]}
// → ValidFrom: nil in DB
// → temporalAnchorHours returns hoursSince(LastAccessed) ≈ 0 instead of hoursSince(2023-01-01)
```

## Affected Code

- `internal/mcp/tools_store.go` lines 300–310 (SHA c409c22)
- Compare with `handleMemoryStore` line 92 (the fix site)

## Proposed Fix

Add `ValidFrom: parseDateTag(itemTags)` to the `types.Memory` literal at line 300, matching `handleMemoryStore` line 92.

Note: `parseDateTag` is package-private in `tools_store.go` — already accessible to `handleMemoryStoreBatch` in the same package.

PR #778, opened 2026-05-20
fix(mcp): store-path divergence cleanup (#762-#765)

petersimmons1972:
## Summary

Fixes the store-path divergence bug class: three sibling MCP handlers diverged from the canonical `handleMemoryStore` fix in #747 (commit `c409c22`) and silently dropped `valid_from`, `episode_id`, or `immutable` fields.

- **#762** — `memory_store_batch` dropped `valid_from`: batch items with `date:` tags received `nil` `ValidFrom`, causing `temporalAnchorHours` to fall back to `LastAccessed` for all bulk-ingested dated memories.
- **#763** — `memory_store_document` dropped both `valid_from` AND `episode_id`: documents stored during a named episode were not linked to that episode; date: tags were silently discarded.
- **#764** — `memory_store_batch` dropped per-item `immutable=true`: the flag was always zero-valued despite the tool docstring promising "same optional fields as memory_store".
- **#765** — `memory_correct` did not recalculate `valid_from` on tag change: correcting a date: tag via `memory_correct` left the stored `valid_from` stale. Edge-case decision: if new tags have **no** date: tag but old tags did, `valid_from` is left unchanged ("only promote, never nullify" — avoids precision loss on partial corrections; documented in commit message and code comment).

## Implementation

- Promoted `parseDateTag` from a private mcp-package function to `types.ParseDateTag` (exported), enabling the db layer to call it without a circular import. The mcp-package shim delegates to it for call-site compatibility.
- Added `valid_from=$N` to both UPDATE branches in `UpdateMemory` (`internal/db/postgres_memory.go`).
- Added `ValidFrom`, `Immutable` to the batch item `types.Memory` literal.
- Added `ValidFrom`, `EpisodeID` to the document `types.Memory` literal.

## Reference

Canonical pattern: #747, commit `c409c22`, `internal/db/postgres_valid_from_test.go`.

## Tests

Four regression tests added to `internal/mcp/tools_store_test.go`:
- `TestMemoryStoreBatch_PersistsValidFrom` (#762)
- `TestMemoryStoreBatch_PersistsImmutable` (#764)
- `TestMemoryStoreDocument_PersistsValidFromAndEpisode` (#763)
- `TestMemoryCorrect_RecalculatesValidFromOnTagChange` (#765)

All pass: `go test ./internal/mcp/... ./internal/db/... ./internal/types/... -count=1 -race`

## Test plan

- [ ] `go build ./...` clean
- [ ] `go vet ./...` clean
- [ ] `go test ./internal/mcp/... -count=1 -race` PASS
- [ ] `go test ./internal/db/... -count=1 -race` PASS (regression check on canonical path)
- [ ] `go test ./internal/types/... -count=1 -race` PASS

### DIFF
--- a/internal/db/postgres_memory.go
+++ b/internal/db/postgres_memory.go
@@ -208,6 +208,13 @@ func (b *PostgresBackend) UpdateMemory(
 	}
 	if tags != nil {
 		m.Tags = tags
+		// Recalculate ValidFrom from the new tags so a date: tag added (or changed)
+		// via memory_correct takes effect. If the new tags have no date: tag but the
+		// old tags did, ValidFrom is left unchanged ("only promote, never nullify"
+		// policy — avoids accidental precision loss on partial corrections). Closes #765.
+		if newVF := types.ParseDateTag(tags); newVF != nil {
+			m.ValidFrom = newVF
+		}
 	}
 	if importance != nil {
 		m.Importance = *importance
@@ -227,13 +234,13 @@ func (b *PostgresBackend) UpdateMemory(
 		m.ContentHash = &hash
 		// Clear the summary so the background worker regenerates it with the new content.
 		_, err = tx.Exec(ctx,
-			"UPDATE memories SET content=$1, tags=$2, importance=$3, updated_at=$4, content_hash=$5, summary=NULL, pattern_confidence=$6 WHERE id=$7 AND project=$8",
-			m.Content, tagsJSON, m.Importance, now, hash, m.PatternConfidence, id, b.project,
+			"UPDATE memories SET content=$1, tags=$2, importance=$3, updated_at=$4, content_hash=$5, summary=NULL, pattern_confidence=$6, valid_from=$7 WHERE id=$8 AND project=$9",
+			m.Content, tagsJSON, m.Importance, now, hash, m.PatternConfidence, m.ValidFrom, id, b.project,
 		)
 	} else {
 		_, err = tx.Exec(ctx,
-			"UPDATE memories SET content=$1, tags=$2, importance=$3, updated_at=$4, pattern_confidence=$5 WHERE id=$6 AND project=$7",
-			m.Content, tagsJSON, m.Importance, now, m.PatternConfidence, id, b.project,
+			"UPDATE memories SET content=$1, tags=$2, importance=$3, updated_at=$4, pattern_confidence=$5, valid_from=$6 WHERE id=$7 AND project=$8",
+			m.Content, tagsJSON, m.Importance, now, m.PatternConfidence, m.ValidFrom, id, b.project,
 		)
 	}
 	if err != nil {
--- a/internal/db/postgres_valid_from_test.go
+++ b/internal/db/postgres_valid_from_test.go
@@ -62,3 +62,66 @@ func TestStoreMemory_NilValidFromStaysNil(t *testing.T) {
 	require.NotNil(t, got)
 	require.Nil(t, got.ValidFrom, "valid_from must remain NULL when not set on the struct")
 }
+
+// TestUpdateMemory_PreservesValidFromWhenTagsHaveNoDate verifies the
+// "only promote, never nullify" policy: if memory_correct sends tags that
+// no longer include a date: tag, the previously-stored valid_from is
+// preserved unchanged. Regression guard for issue #765.
+func TestUpdateMemory_PreservesValidFromWhenTagsHaveNoDate(t *testing.T) {
+	proj := uniqueProject("update-vf-preserve")
+	b := newTestBackend(t, proj)
+	ctx := context.Background()
+
+	// 1. Store a memory with a date: tag — ValidFrom should be set.
+	originalDate := time.Date(2024, 6, 15, 0, 0, 0, 0, time.UTC)
+	m := &types.Memory{
+		ID:         types.NewMemoryID(),
+		Content:    "test content for preserve",
+		Project:    proj,
+		MemoryType: types.MemoryTypeContext,
+		Importance: 1,
+		Tags:       []string{"date:2024-06-15", "foo"},
+		ValidFrom:  &originalDate,
+	}
+	require.NoError(t, b.StoreMemory(ctx, m))
+
+	// 2. Call UpdateMemory with tags that have NO date: tag.
+	newTags := []string{"foo", "bar"}
+	updated, err := b.UpdateMemory(ctx, m.ID, nil, newTags, nil, nil)
+	require.NoError(t, err)
+
+	// 3. ValidFrom must still equal the original date — preserved, not cleared.
+	require.NotNil(t, updated.ValidFrom, "ValidFrom must not be nullified when new tags omit date:")
+	require.True(t, updated.ValidFrom.Equal(originalDate),
+		"ValidFrom must equal original 2024-06-15, got %s — see issue #765 'only promote, never nullify'", updated.ValidFrom)
+}
+
+// TestUpdateMemory_PromotesValidFromOnDateTagChange is the paired positive case:
+// when new tags include a different date: tag, ValidFrom is updated to match.
+func TestUpdateMemory_PromotesValidFromOnDateTagChange(t *testing.T) {
+	proj := uniqueProject("update-vf-promote")
+	b := newTestBackend(t, proj)
+	ctx := context.Background()
+
+	originalDate := time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC)
+	m := &types.Memory{
+		ID:         types.NewMemoryID(),
+		Content:    "test content for promote",
+		Project:    proj,
+		MemoryType: types.MemoryTypeContext,
+		Importance: 1,
+		Tags:       []string{"date:2024-01-01"},
+		ValidFrom:  &originalDate,
+	}
+	require.NoError(t, b.StoreMemory(ctx, m))
+
+	// Update with a new date: tag — ValidFrom must be promoted to the new date.
+	newTags := []string{"date:2024-12-31"}
+	updated, err := b.UpdateMemory(ctx, m.ID, nil, newTags, nil, nil)
+	require.NoError(t, err)
+
+	expectedNew := time.Date(2024, 12, 31, 0, 0, 0, 0, time.UTC)
+	require.NotNil(t, updated.ValidFrom)
+	require.True(t, updated.ValidFrom.Equal(expectedNew),
+		"ValidFrom must equal new 2024-12-31, got %s", updated.ValidFrom)
+}
--- a/internal/mcp/tools_store.go
+++ b/internal/mcp/tools_store.go
@@ -189,6 +189,8 @@ func handleMemoryStoreDocument(ctx context.Context, pool *EnginePool, req mcpgo.
 		Importance: importance,
 		Tags:       docTags,
 		Immutable:  getBool(args, "immutable", false),
+		EpisodeID:  episodeIDFromContextOrArgs(ctx, args),
+		ValidFrom:  parseDateTag(docTags),
 	}
 	engine := h.Engine
 	deps := storeDocumentDeps{
@@ -304,8 +306,10 @@ func handleMemoryStoreBatch(ctx context.Context, pool *EnginePool, req mcpgo.Cal
 			Project:           project,
 			Importance:        importance,
 			Tags:              itemTags,
+			Immutable:         getBool(mmap, "immutable", false),
 			StorageMode:       "focused",
 			EpisodeID:         itemEpisodeID,
+			ValidFrom:         parseDateTag(itemTags),
 			PatternConfidence: itemPatternConfidence,
 		})
 	}
@@ -504,31 +508,11 @@ func handleMemoryQuickStore(ctx context.Context, pool *EnginePool, req mcpgo.Cal
 	return handleMemoryStore(ctx, pool, req2, cfg)
 }
 
-// parseDateTag scans tags for a "date:<value>" entry and returns the parsed
-// UTC time (date portion only), or nil if no valid date tag is found.
-// The first matching tag wins. Supported formats:
-//   - "2006-01-02"                — ISO date
-//   - "2006/01/02 (Mon) 15:04"   — LongMemEval haystack date format
-//
-// This is used to populate ValidFrom so the recency scorer uses the actual
-// session date rather than the arbitrary ingest/access time.
+// parseDateTag is a package-local shim that delegates to types.ParseDateTag.
+// Kept for call-site compatibility within this file; callers in other packages
+// (e.g., internal/db) import types.ParseDateTag directly.
 func parseDateTag(tags []string) *time.Time {
-	layouts := []string{"2006-01-02", "2006/01/02 (Mon) 15:04"}
-	for _, tag := range tags {
-		val, ok := strings.CutPrefix(tag, "date:")
-		if !ok {
-			continue
-		}
-		for _, layout := range layouts {
-			t, err := time.Parse(layout, val)
-			if err != nil {
-				continue
-			}
-			t = time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
-			return &t
-		}
-	}
-	return nil
+	return types.ParseDateTag(tags)
 }
 
 // handleMemoryQuery is a simplified front door for handleMemoryRecall.
--- a/internal/mcp/tools_store_test.go
+++ b/internal/mcp/tools_store_test.go
@@ -1,0 +1,227 @@
+package mcp
+
+// Regression tests for the store-path divergence bug class (issues #762–#765).
+//
+// These four tests mirror the canonical regression guard in
+// internal/db/postgres_valid_from_test.go but exercise the MCP handler layer
+// rather than the DB layer, so they work without a Postgres instance.
+//
+// All tests are in-package (no _test suffix) and reuse helpers from
+// simple_tools_test.go (capturingBackend / newCapturingPool) and
+// tools_store_pattern_confidence_test.go (pcCorrectBackend / newPCCorrectPool).
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	mcpgo "github.com/mark3labs/mcp-go/mcp"
+	"github.com/petersimmons1972/engram/internal/db"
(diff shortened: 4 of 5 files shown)


petersimmons1972:

---

#### #762 — `memory_store_batch` `ValidFrom` (PASS)

`parseDateTag(itemTags)` is called after `itemTags` is assembled from `mmap["tags"]` and before the `types.Memory` literal is appended. Call site is correct. The source map is `mmap` (the per-item object), not the outer `args` — correct.

#### #763 — `memory_store_document` `ValidFrom` + `EpisodeID` (PASS)

Both fields are present in the `types.Memory` literal:
- `EpisodeID: episodeIDFromContextOrArgs(ctx, args)` — mirrors the canonical `handleMemoryStore` exactly.
- `ValidFrom: parseDateTag(docTags)` — `docTags` is assembled immediately above from `args["tags"]`, so the tags slice is final at call time.

#### #764 — `memory_store_batch` per-item `immutable` (PASS)

`getBool(mmap, "immutable", false)` — `mmap` is the per-item map object, not the outer batch args, so immutability is per-item as documented. Matches canonical handler's `getBool(args, "immutable", false)` pattern exactly.

#### #765 — `UpdateMemory` `valid_from` recalc (PASS WITH NIT)

Both UPDATE branches (`content != nil` and `else`) now include `valid_from=$N` with `m.ValidFrom` as the parameter. The "only promote, never nullify" guard (`if newVF := types.ParseDateTag(tags); newVF != nil`) is applied before the branches, so `m.ValidFrom` retains its pre-correction value when new tags lack a `date:` tag. This is consistent with the stated intent.

**Nit (nice-to-have):** When `tags == nil` (caller did not pass tags at all), the `valid_from` column is still updated to `m.ValidFrom` — which is the existing DB value read at the start of the function. This is benign (a no-op write), but it's a silent extra column touch in both branches. Not a bug; documenting for awareness.

#### `types.ParseDateTag` extraction — line-by-line equivalence (PASS)

The exported function in `internal/types/types.go` is a byte-for-byte copy of the removed implementation from `tools_store.go`. The shim in `tools_store.go` delegates correctly.

#### Nil-deref / empty input safety (PASS)

`ParseDateTag` iterates `tags`; an empty or nil slice causes the loop to produce zero iterations and return `nil`. No panic path. `getBool(mmap, ...)` with a nil `mmap` would panic, but `mmap` is guarded by the `item.(map[string]any)` type assertion immediately above.

#### Test invertibility — do the tests fail before the fix? (PASS WITH NIT)

- **#762 / #764:** `capturingBackend` captures via `StoreMemoryTx`, and `StoreBatch` calls `StoreMemoryTx` per item inside the transaction. Capture path is valid. Reverting the `ValidFrom`/`Immutable` fields from the `types.Memory` literal would make both tests fail — genuine regression guards.
- **#763:** `documentCapturingBackend` overrides `StoreMemoryTx`. `execStoreDocument` routes through that path. Reverting would make the test fail.
- **#765 (NIT — important):** `TestMemoryCorrect_RecalculatesValidFromOnTagChange` verifies `back.capturedTags` contains `"date:2024-12-31"` and that `ParseDateTag(back.capturedTags)` returns the expected time. However, the test backend's `UpdateMemory` implementation calls `types.ParseDateTag(tags)` itself to simulate the real DB — so the test passes whether or not `UpdateMemory` in `postgres_memory.go` actually recalculates `valid_from`. **The MCP-layer test cannot catch a regression in the DB-layer SQL.** This is structural (not a defect in the fix itself), and `internal/db/postgres_valid_from_test.go` is cited as the canonical regression guard for the DB layer. Acceptable, but callers reading the test file may incorrectly believe the SQL is covered here. A one-line comment cross-referencing the DB test file would prevent confusion.

---

#### Summary

| Issue | Verdict |
|-------|---------|
| #762 batch ValidFrom | PASS |
| #763 document ValidFrom + EpisodeID | PASS |
| #764 per-item immutable | PASS |
| #765 UpdateMemory recalc | PASS |
| ParseDateTag extraction equivalence | PASS |
| Nil-deref / empty input | PASS |
| Test invertibility | PASS with nit on #765 test scope |

**Blockers:** 0

**Nits (severity/nice-to-have):**
1. `#765` test comment should cross-reference `internal/db/postgres_valid_from_test.go` as the SQL regression guard so future readers understand the test boundary.
2. Both UPDATE branches write `valid_from` even when tags are nil (benign no-op write; document or suppress if it causes query plan concerns).

**APPROVE-WITH-NITS** — safe to proceed to Round 2 (Coverage).

petersimmons1972:

**Artifact**: PR #778 — store-path divergence cleanup (#762–#765), 4 files, +276/-28

---

#### Trust Breakpoints

1. **Location**: `internal/types/types.go:409–436` — `ParseDateTag` placement
   **Observation**: `package types` opens with "core data structures... All values are ported from the Python reference implementation... must remain wire-compatible with it." `ParseDateTag` is operational parsing logic — it classifies a tag string against two date layouts. No Python wire-format has a "date:" tag parsing convention that this needs to mirror. It is not a data structure, a constant, or a wire-format value.
   **Concern**: The placement was driven by import graph necessity (circular import avoidance), not semantic fit. `internal/types` is the repo's most-imported package; loading it pulls `ParseDateTag` into every consumer, including packages that will never call it. More critically, if the date-tag format ever needs to change or extend, the change now sits inside the wire-compat package — raising the (false) concern that it's a wire-format change. An `internal/tag` or `internal/parse` leaf package would have been semantically honest and equally import-cycle-free. That said, this is not a correctness defect, and the function is small and pure.
   **Severity**: nitpick

2. **Location**: `internal/mcp/tools_store.go:511–515` — `parseDateTag` shim retained
   **Observation**: The shim is a one-liner delegating to `types.ParseDateTag`. There are three call sites in the same file (lines 92, 193, 312). All three could call `types.ParseDateTag` directly with no behavioral change. The shim comment says it is "kept for call-site compatibility within this file."
   **Concern**: The stated justification does not hold — intra-file call sites have no compatibility constraint; they can be updated atomically in the same PR. The shim now means `date_tag_test.go` tests the shim, not the canonical function. `types.ParseDateTag` has zero dedicated tests (confirmed: `internal/types/types_test.go` contains none). The only test coverage of `ParseDateTag`'s logic runs via the shim, which is an indirection layer away from the actual export. If the shim were ever removed and call sites updated, the tests in `date_tag_test.go` would break — not because the logic changed but because the shim disappeared. The cleaner outcome is: remove the shim, update the three call sites, migrate `date_tag_test.go` to `internal/types/types_test.go` (or a new `internal/types/parse_test.go`). The PR chose the safer/lazier path.
   **Severity**: nitpick

3. **Location**: `internal/db/postgres_memory.go:232–244` — two-branch SQL with `valid_from` in both
   **Observation**: The content-changed branch (line 237) and the tags-only branch (line 242) both now include `valid_from=$N`. The positional parameter numbering was correctly updated in both branches ($7 in the tags-only branch, $7/$8/$9 in the content branch).
   **Concern**: The two branches are now structurally identical except for `content_hash`, `summary=NULL`, and parameter count. There is no factoring that prevents a future change from updating one branch and missing the other — the "only promote, never nullify" recalculation at line 215 happens before both branches and is shared, but the SQL strings are duplicated. This is the canonical copy-paste drift smell. Not a defect today; a defect waiting to happen if a third field gets added. Would flag for a TODO comment at minimum.
   **Severity**: nitpick

4. **Location**: `internal/db/postgres_memory.go:209–217` — "only promote, never nullify" policy
   **Observation**: If `tags != nil` but the new tags contain no `date:` entry, `ValidFrom` is left unchanged at whatever the prior value was. This is documented in the comment and PR body. The policy is intentional.
   **Concern**: There is no mechanism for a caller to express "I explicitly want to clear `valid_from`." The `memory_correct` tool accepts `tags` as a plain `[]string` with no sentinel value. A caller who wants to strip a date tag must send the full desired tag list without any `date:` entry — but the "only promote, never nullify" logic will silently ignore that intent and preserve the old `valid_from`. This is a closed door, not a bug in the current behavior, but the policy decision is not surfaced to the tool's callers in any way. There is no tool-level documentation saying "removing a date: tag from tags does not clear valid_from." That is a user-facing correctness trap if anyone relies on tag correction to retract a date.
   **Severity**: serious — not a blocker because the behavior is intentional and documented in the code, but it should either be surfaced in the tool docstring or a follow-up issue should be filed before merge. A caller who reads only the tool description will not know this invariant exists.

5. **Location**: `internal/mcp/tools_store_test.go` — new test file naming
   **Observation**: The file is `tools_store_test.go` in `package mcp`. Cross-checked against existing files — `tools_store_pattern_confidence_test.go`, `tools_store_batch_pattern_confidence_test.go`, `tools_store_decouple_test.go`, `tools_store_permanent_test.go` all exist. No collision. The name is consistent with the naming convention in this package.
   **Concern**: None.
   **Severity**: (no issue)

6. **Location**: `internal/types` — `ParseDateTag` export surface
   **Observation**: `internal/` packages in Go are not importable outside the module, so no external callers can be affected. Confirmed the module path is `github.com/petersimmons1972/engram` and all callers are within that module.
   **Concern**: None re: external breakage.
   **Severity**: (no issue)

---

#### What Earned Trust

- The `valid_from` recalculation in `UpdateMemory` is positioned correctly — after `m.Tags = tags` so the recalculation operates on the new tag set, not the old one. That ordering is not accidental and whoever wrote it thought about it.
- Both SQL branches were updated and parameter ordinals were renumbered without error. Getting positional parameter counts wrong in multi-branch SQL is a common class of defect; it was avoided here.
- Regression test #765 (`TestMemoryCorrect_RecalculatesValidFromOnTagChange`) correctly uses the simulated backend to verify the tag reaches `UpdateMemory` — it does not try to mock the recalculation logic, which would have been an overfit test. The test boundary is appropriate.
- `documentCapturingBackend` is a purpose-built struct that avoids polluting the general `capturingBackend` with `StoreMemoryTx` capture. Shows awareness of test isolation.

---

1. **"Only promote, never nullify" — is this in the tool docstring?** If a caller sends `tags: ["foo", "bar"]` (no `date:` tag) to `memory_correct`, their prior `valid_from` is silently preserved. Is this behavior documented anywhere the tool consumer can see before they make the call?

2. **`types.ParseDateTag` coverage gap** — `date_tag_test.go` tests the shim, which means removing the shim later would break those tests mechanically. Is there a plan to migrate those tests to `internal/types/types_test.go`, or is the shim considered permanent?

3. **Two-branch SQL drift** — is there a TODO, issue, or comment tracking the intention to consolidate the two `UPDATE` branches in `UpdateMemory` before a third field gets added?

4. **Shim justification** — the comment says the shim is kept for "call-site compatibility within this file." There are only three intra-file call sites. What is the compatibility constraint that prevents updating them in this PR?

---

**Verdict**: APPROVE-WITH-NITS. Zero blockers. Finding #4 (nullify path silently closed) is serious and should be resolved before merge — either a tool-docstring update or a filed tracking issue. Findings #1–#3 are nitpicks that do not block.

petersimmons1972:

**Coverage** (without `TEST_DATABASE_URL` — integration tests skipped per canonical pattern):
- `internal/mcp`: **63.9%** ✅ (above 60% target; above 55% CI gate)
- `internal/types` (package-isolated): **48.9%** ⚠️ — below 60% target, but `ParseDateTag` itself hits **100%** when measured cross-package via mcp test suite (`-coverpkg=./...`)
- `internal/mcp/tools_store.go` per-function: `handleMemoryStore` 76.8% · `handleMemoryStoreDocument` 76.3% · `handleMemoryStoreBatch` 75.0% · `handleMemoryCorrect` 86.5% · `parseDateTag` shim 100%

---

**All four new tests pass** — confirmed with `-count=1 -race`.

---

#### Findings

**[severity/nice-to-have] #1 — TestMemoryStoreBatch_PersistsValidFrom: missing zero-tags + malformed-date-tag boundary variants**

Policy (CLAUDE.md Test Policy): "New MCP tool handlers require at minimum: happy path, zero/empty input, and one boundary condition test."

Missing:
- Batch item with no tags → `ValidFrom` stays `nil` (zero-input variant)
- Batch item with `date:not-a-date` tag → `ValidFrom` stays `nil`, no panic (boundary variant)

Not a blocker because the `parseDateTag` shim and `types.ParseDateTag` are already boundary-tested in `internal/mcp/date_tag_test.go` (nil, invalid, none-present, empty — all covered). The regression risk is low, but the per-handler zero/empty contract is unverified at the handler layer.

**[severity/nice-to-have] #2 — TestMemoryStoreBatch_PersistsImmutable: missing `immutable: false` default-off variant**

The default-off case (`immutable` key absent from args → `false`) is not explicitly tested at the handler layer. `getBool(mmap, "immutable", false)` returning false for a missing key is a mechanical transform, but the policy calls for it. Low risk given the trivial logic.

**[severity/nice-to-have] #3 — TestMemoryStoreDocument_PersistsValidFromAndEpisode: missing neither-default variant**

No test for: content with no date tag and no episode context → `ValidFrom == nil`, `EpisodeID == ""`. The happy path documents the positive case; the default-fallback case is untested at the handler layer.

**[severity/blocker] #4 — TestMemoryCorrect_RecalculatesValidFromOnTagChange: "only promote, never nullify" policy has no regression test**

The PR description and `postgres_memory.go` comment explicitly document a "never nullify" edge case: if new tags have no `date:` tag but old tags did, `ValidFrom` is left unchanged. This is a load-bearing design decision that closes half the correctness contract for `memory_correct`. There is **no test** that sends tags-without-date to `handleMemoryCorrect` and asserts the old `ValidFrom` is preserved.

The `validFromCorrectBackend.UpdateMemory` stub simulates the DB recalculation via `types.ParseDateTag(tags)` — which correctly returns `nil` for no-date tags — but the test never calls `handleMemoryCorrect` with no-date tags. The documented invariant is not regression-guarded at any layer. If a future refactor changes the `if newVF := ...; newVF != nil` guard in `UpdateMemory` to unconditional assignment, no test will catch it.

**[severity/nice-to-have] #5 — `types.ParseDateTag` has 0% coverage in package-isolated `internal/types` profile**

Package-isolated `go test ./internal/types/...` reports 48.9% with `ParseDateTag` at 0% (no test in the types package exercises it directly). Cross-package measurement (`-coverpkg=./...`) shows 100% via mcp tests. The function is adequately exercised; this is a reporting gap, not a functional gap. A direct test in `internal/types/types_test.go` would make the types package self-contained and prevent the coverage from regressing if mcp tests are reorganized.

---

#### Verdict

**APPROVE-WITH-NITS** — one blocker (#4): the "only promote, never nullify" policy is explicitly documented but has no regression test at any layer. A minimal test sending no-date tags to `handleMemoryCorrect` with a backend that records whether `ValidFrom` changed would close this gap. Items #1–#3 and #5 are `severity/nice-to-have` and do not block merge.

petersimmons1972:

**Finding 1 — `severity/nice-to-have` — #765 regression test does not exercise the actual fix.**
`TestMemoryCorrect_RecalculatesValidFromOnTagChange` uses `validFromCorrectBackend.UpdateMemory`, a hand-rolled fake that *itself* calls `types.ParseDateTag(tags)` to populate `ValidFrom` on the returned memory. The real fix lives in `internal/db/postgres_memory.go` `UpdateMemory`. The MCP-layer test therefore only proves that the new `tags` slice reaches `UpdateMemory` — which was never the bug. The final assertion (`gotVF := types.ParseDateTag(back.capturedTags); require.True(gotVF.Equal(want))`) is tautological: it parses the tag the test itself just constructed. The real "only promote, never nullify" semantics (the load-bearing behavioral decision in the PR) are not covered here. Recommend either (a) adding a DB-layer test in `postgres_valid_from_test.go` that exercises a real UPDATE with a nullify-attempt and asserts `valid_from` is preserved, or (b) referencing an existing canonical test that covers the same path.

**Finding 2 — `severity/nice-to-have` — `parseDateTag` shim is dead weight if no in-package callers remain.**
The PR retains `func parseDateTag(tags []string) *time.Time { return types.ParseDateTag(tags) }` in `tools_store.go` "for call-site compatibility." If the only remaining callers are the two new lines added in this same PR, the shim provides no migration value. Either inline it (preferred) or leave a comment naming the callers that justify keeping it.

**Finding 3 — `severity/nice-to-have` — "only promote, never nullify" policy is asymmetric with other update semantics.**
In `UpdateMemory`, `content` and `importance` are conditional-update; `tags` is a full replacement. The new ValidFrom logic adds a third semantic: replace-but-only-if-better. Defensible, but it means `memory_correct` can leave `valid_from` and `tags` in disagreement. Worth a brief docstring on `memory_correct` so future callers don't trip on it.

**Finding 4 — `severity/nice-to-have` — confirm the batch path persists `Immutable` end-to-end.**
The new test asserts `cap.stored[0].Immutable == true` at the capturing-backend layer. Worth confirming the engine's bulk-store path forwards `Immutable` to the DB column — the whole point of this PR is that "looks set at the handler" was an insufficient guarantee. A quick grep on `internal/search/` and `internal/db/` would close the loop.

**Finding 5 — `nit` — commit/PR body promises "documented in commit message and code comment."**
The code comment in `postgres_memory.go` is present and clear. Verify the commit message carries the same rationale so `git log` readers don't have to chase the PR.

**Verdict: APPROVE-WITH-NITS — zero blockers. Fix is correct, narrow, and the shape (canonical impl in `types`, shim in `mcp`) is the right call to avoid the circular import. File the strongest one (#765 test depth) as a follow-up issue per severity-gating policy.**

petersimmons1972:
### Blocker fix applied — only-promote-never-nullify regression coverage

Added two paired tests in \`internal/db/postgres_valid_from_test.go\` (commit bc2a7b5):

- \`TestUpdateMemory_PreservesValidFromWhenTagsHaveNoDate\` — stores a memory with \`date:2024-06-15\`, then calls \`UpdateMemory\` with tags that have no \`date:\` tag, asserts \`ValidFrom\` is still \`2024-06-15\`. Direct regression guard for the "only promote, never nullify" policy at \`postgres_memory.go:215-217\`.
- \`TestUpdateMemory_PromotesValidFromOnDateTagChange\` — paired positive case verifying date promotion still works when a new \`date:\` tag is present.

Both tests skip without \`TEST_DATABASE_URL\` (canonical pattern via \`testutil.DSN\`). CI: Lint ✅ · Test & Coverage ✅ · Secret scan ✅ · GitGuardian ✅ · Snyk ❌ (quota exhausted, pre-existing).

\`severity/blocker\` removed. Re-review of Round 2 (Coverage) requested.

PR #786, opened 2026-05-20
fix(mcp): narrow importance bounds to 0-4 (#768)

### LINKED ISSUE #768: validate importance bounds consistently across MCP store paths
tools_store.go validateQuickStoreInput accepts importance 0–100 with named constants maxImportanceValue/minImportanceValue, but handleMemoryStore rejects anything outside 0–4. Callers passing the doc'd 0–100 range get a confusing downstream error. Either widen handleMemoryStore, narrow the REST validator, or add a documented clamp. Discovered during PR #766 Round 3 review.

petersimmons1972:
## Summary

Fixes #768 — importance bounds inconsistency between `validateQuickStoreInput` and `handleMemoryStore`.

**Decision:** Narrow the REST validator to 0-4 (match the handler), not widen the handler. The handler is the source of truth on semantic meaning; widening it would silently invalidate existing stored data.

## What changed

| File | Change |
|------|--------|
| `internal/mcp/tools_store.go` | `maxImportanceValue`: 100 to 4; constant comment and `validateQuickStoreInput` docstring updated to say 0-4 |
| `internal/mcp/quick_store_handler_test.go` | `TestQuickStoreHandler_InvalidImportance` boundary values updated from `{-1,101}` to `{-1,5}` (old values encoded the bug); added `TestValidateQuickStoreInput_RejectsImportanceAbove4` regression test covering values 5, 10, 50, 100 |
| `docs/architecture.md` | Corrected "Importance: 0-100" bullet to "0-4 (matches the handler range)" |

## Test plan

- [x] `go build ./internal/mcp/...` — clean
- [x] `go vet ./...` — clean (pre-existing build failures in internal/cache, internal/runner, internal/reporter are unrelated to this change)
- [x] `go test ./internal/mcp/... -count=1 -race` — all pass (15s)
- [x] `TestValidateQuickStoreInput_RejectsImportanceAbove4` — new regression test, covers 5/10/50/100
- [x] `TestQuickStoreHandler_InvalidImportance` — updated to reflect correct 0-4 boundary

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -127,7 +127,7 @@ The `/quick-store` and `/quick-recall` REST endpoints enforce per-field limits:
 - **Content Size:** max 1 MiB (quick-store only)
 - **Project Name:** must match `^[a-z0-9_-]{1,64}$` (alphanumeric, underscore, hyphen; no spaces, uppercase, or path traversal)
 - **Tags:** max 64 total; each max 256 characters
-- **Importance:** 0–100 (not 0–4 like memory_type)
+- **Importance:** 0–4 (matches the handler range; 0=Critical, 4=Low)
 - **Query:** must be non-empty
 
 Validation failures return HTTP 400 with structured error message.
--- a/internal/mcp/quick_store_handler_test.go
+++ b/internal/mcp/quick_store_handler_test.go
@@ -182,9 +182,11 @@ func TestQuickStoreHandler_TagTooLong(t *testing.T) {
 	require.Equal(t, http.StatusBadRequest, w.Code)
 }
 
-// TestQuickStoreHandler_InvalidImportance verifies that importance outside [0,100] is rejected with 400.
+// TestQuickStoreHandler_InvalidImportance verifies that importance outside [0,4] is rejected with 400.
+// The range was previously documented as 0–100 but the handler (handleMemoryStore) always enforced
+// 0–4. This test was updated as part of fix #768 to match the narrowed validator.
 func TestQuickStoreHandler_InvalidImportance(t *testing.T) {
-	tests := []int{-1, 101}
+	tests := []int{-1, 5}
 	for _, imp := range tests {
 		t.Run(fmt.Sprintf("importance=%d", imp), func(t *testing.T) {
 			s := newQuickStoreServer(t)
@@ -201,6 +203,19 @@ func TestQuickStoreHandler_InvalidImportance(t *testing.T) {
 			s.handleQuickStore(w, req)
 
 			require.Equal(t, http.StatusBadRequest, w.Code)
+		})
+	}
+}
+
+// TestValidateQuickStoreInput_RejectsImportanceAbove4 is the regression test for #768.
+// Values in the formerly-accepted range (5–100) must now be rejected by the validator
+// before they reach handleMemoryStore, giving callers an early, descriptive error.
+func TestValidateQuickStoreInput_RejectsImportanceAbove4(t *testing.T) {
+	for _, imp := range []int{5, 10, 50, 100} {
+		t.Run(fmt.Sprintf("importance=%d", imp), func(t *testing.T) {
+			err := validateQuickStoreInput("some content", "global", nil, imp)
+			require.Error(t, err, "importance=%d should be rejected — was accepted by the buggy 0-100 validator", imp)
+			require.Contains(t, err.Error(), "importance must be", "error message should describe the valid range")
 		})
 	}
 }
--- a/internal/mcp/tools_store.go
+++ b/internal/mcp/tools_store.go
@@ -441,7 +441,8 @@ const maxQuickStoreTags = 64
 const maxQuickStoreTagLength = 256
 
 // maxImportanceValue is the maximum allowed importance value.
-const maxImportanceValue = 100
+// Must match the 0–4 range enforced by handleMemoryStore.
+const maxImportanceValue = 4
 
 // minImportanceValue is the minimum allowed importance value.
 const minImportanceValue = 0
@@ -453,7 +454,7 @@ var projectNamePattern = regexp.MustCompile(`^[a-z0-9_-]{1,64}$`)
 // - content: required, max 1 MiB
 // - project: must match ^[a-z0-9_-]{1,64}$
 // - tags: max 64, each max 256 chars
-// - importance: 0–100.
+// - importance: 0–4.
 func validateQuickStoreInput(content string, project string, tags []string, importance int) error {
 	if len(content) > maxQuickStoreContentSize {
 		return fmt.Errorf("content exceeds max size %d bytes", maxQuickStoreContentSize)


petersimmons1972:

**Scope verified:** 3 files, ~21 LOC, narrowing `maxImportanceValue` from 100 to 4.

---

#### Findings

**[nice-to-have] `TestQuickStoreHandler_InvalidImportance` boundary set is sparse**

The updated test exercises `{-1, 5}`. Value `5` is the first out-of-bounds value, which is good. However, the old boundary was `101` — a value one past the old limit. Symmetry with the new limit would add `5` (already present) _and_ confirm `4` is accepted. The valid upper-bound acceptance case (`importance=4`) is not explicitly tested in this test function (it is implicitly covered by the regression test sampling `5, 10, 50, 100`, but not via the HTTP handler path). Not a blocker; flagged for completeness.

**[nice-to-have] `score.go` clamping comment references `importance>=5` as a "hand-inserted row" scenario**

`internal/search/score.go:272–275` still contains a comment noting that `importance>=5` can arrive from "hand-inserted rows" and that the 0.1 floor prevents a zero ranking. This comment is accurate (the DB has no enforcement — only the API layer does), but after this PR the comment is the only documentation of that escape hatch. No change needed in this PR; recommend a follow-up issue to note the intentional DB-layer gap.

**[no-issue] Migration story is clean**

The fix is write-path only. Existing stored memories with importance values > 4 are not touched (no schema migration, no backfill). The `ImportanceBoost` function in `score.go` already handles out-of-range values via the 0.1 floor clamp. No data integrity risk.

**[no-issue] No other consumers of the old 0–100 bound found**

`grep -rn maxImportanceValue internal/` shows the constant is used exclusively in `validateQuickStoreInput`. All other handlers (`handleMemoryStore`, `handleMemoryStoreBatch`, `handleDocumentStore`) independently validate against the literal `4` — the new constant comment ("Must match the 0–4 range enforced by handleMemoryStore") correctly documents this coupling.

**[no-issue] Doc fix is accurate**

`docs/architecture.md` now reads `0–4 (matches the handler range; 0=Critical, 4=Low)`. The semantic labels match the `score.go` docstring (`importance=0 → critical`, `importance=4 → trivial`). Consistent.

---

#### Verdict

**APPROVE-WITH-NITS.** Zero blockers. Two nits, both nice-to-have. The logic is correct, the constant comment explicitly documents the coupling to `handleMemoryStore`, and the regression test covers the full formerly-accepted range (5, 10, 50, 100). The existing data migration story is sound.

petersimmons1972:

**1. [severity/nice-to-have] Boundary test update — old values undocumented**
The pre-existing test had boundary values `{-1, 101}`. They are now `{-1, 5}`. The arithmetic is clear (one above each bound). Verify the test *name* and *comments* no longer reference the old `101`/100 numbers anywhere, to avoid future confusion.

**2. [severity/nice-to-have] New test name asymmetry**
The new test `TestValidateQuickStoreInput_RejectsImportanceAbove4` baked the bound into the name. If the bound later changes, the test name becomes a lie. Consider `TestValidateQuickStoreInput_RejectsImportanceAboveMax` and assert against the constant.

**3. [severity/nice-to-have] `docs/architecture.md` — verify all occurrences**
If the importance range appears in multiple places in `architecture.md` (definition, example, constraints table), confirm all were updated. A stale "0–100" in a comment or example row is the exact kind of thing that causes the next contributor to re-introduce the wrong constant.

**4. [severity/nice-to-have] Missing negative-bound test**
The fix narrative focuses on the upper bound (`100 → 4`). New test covers `>4`. The summary does not mention a corresponding `<0` rejection test. If that coverage exists elsewhere, disregard.

**5. [severity/nice-to-have] No changelog or migration note**
Consumers sending importance values in `5–100` today will now receive a validator-level rejection instead of a handler-level rejection. Error message and HTTP status code may differ. A one-line changelog entry would help integrators.

---

**Verdict: APPROVE-WITH-NITS — 0 blockers.** Core change is correct and well-scoped. Nits are documentation/test hygiene.

petersimmons1972:

**Coverage:** `validateQuickStoreInput` 100.0% | `handleMemoryQuickStore` 100.0% | `internal/mcp` package 63.9% (above 60% target gate)

---

**Findings:**

`[severity/nice-to-have]` **Missing boundary-low acceptance tests for valid range (0, 2, 3, 4).**
The test policy requires "happy path + boundary." The happy-path tests that exist use `importance=1` only (lines 52, 154, 174). No test explicitly verifies that `0` (Critical) and `4` (Low) — the exact boundary values — are accepted by `validateQuickStoreInput`. A future change to the bounds could silently break these without a failing test. Recommend adding a `TestValidateQuickStoreInput_AcceptsValidRange` covering `[]int{0, 1, 2, 3, 4}`.

`[severity/nice-to-have]` **`TestValidateQuickStoreInput_RejectsImportanceAbove4` does not cover below-bound (`-1`).**
Below-bound rejection is covered only through the HTTP-layer `TestQuickStoreHandler_InvalidImportance` (which does include `-1, 5`). That's sufficient for correctness, but the regression test by name implies exhaustive validator coverage — consider adding `-1` to its table or noting the split in the comment.

---

**Summary:** No blockers. Both rejection directions are covered (below via `TestQuickStoreHandler_InvalidImportance{-1,5}`, above via `TestValidateQuickStoreInput_RejectsImportanceAbove4{5,10,50,100}`). The only gap is explicit acceptance tests for boundary values 0 and 4, which is a nice-to-have per the test policy's "happy + boundary" requirement.

**Verdict: APPROVE-WITH-NITS** — 0 blockers, 2 nice-to-haves. Safe to merge; acceptance boundary tests recommended as follow-on.

petersimmons1972:

**Artifact**: PR #786 — fix(mcp): narrow importance bounds to 0-4 (#768)

---

### Trust Breakpoints

1. **Location**: `internal/mcp/tools_store.go:376–379` (`handleMemoryCorrect`)
   **Observation**: When importance is provided to `memory_correct`, the value goes through `types.ValidateImportance()` which **silently clamps** to [0, 4] rather than returning an error. A caller passing `importance=50` gets it stored as `4` with no feedback. This is the same class of silent-corruption that `ValidatePatternConfidence` was deliberately redesigned to avoid — the comment at `types.go:117–119` explicitly calls out this difference as a known tradeoff. The PR tightens `validateQuickStoreInput` (REST path) and `handleMemoryStore` already errors on out-of-range, but `handleMemoryCorrect` still silently clamps.
   **Concern**: A REST client that was previously sending `importance=50` via `memory_correct` will continue to receive a success response with the value silently truncated to `4`. This is not a regression introduced by this PR, but this PR's stated goal is to surface the boundary clearly — and it only does so on the store path, not the correct path. The architectural asymmetry is now more visible and more confusing.
   **Severity**: serious

2. **Location**: `maxImportanceValue` constant name in `internal/mcp/tools_store.go:443`
   **Observation**: The constant is named `maxImportanceValue` and its value is `4`. In the domain, `0` is Critical (highest priority) and `4` is Trivial/Low (lowest priority). The name `maxImportanceValue` reads as "the most important level" but semantically it is "the least important level" — the highest integer in an inverted scale.
   **Concern**: A future maintainer reading `importance < minImportanceValue || importance > maxImportanceValue` will likely parse `maxImportanceValue` as the most-important bound. This is a landmine for the next person who touches the pruning logic. `maxImportanceRank` or `lowestImportanceLevel` would be unambiguous.
   **Severity**: nitpick

3. **Location**: `docs/tools.md` (importance table, lines ~104–114) and `internal/mcp/server.go:1266` (memory_correct description string)
   **Observation**: `docs/tools.md` has the correct 0-4 table with named levels. `server.go:1266` describes `memory_correct` as accepting importance but does not specify range. No stale 0-100 references found in user-facing docs or non-archived plan files — the doc sweep was thorough. This is a positive signal, not a concern.
   **Concern**: None beyond the nitpick that the MCP tool description for `memory_correct` is silent on the valid range while `memory_store`'s error message at least says "0–4" on rejection.
   **Severity**: nitpick

4. **Location**: PR decision rationale — "widening the handler would silently invalidate existing stored data"
   **Observation**: The PR body claims widening `handleMemoryStore` to accept 0-100 would invalidate existing stored data. This claim is not substantiated. The multiplier formula is `(5 - importance) / 3.0` — values above 4 would produce a multiplier below `1/3`, but nothing in the codebase currently uses values above 4. Existing stored data would be unaffected; only future stores would use the extended range.
   **Concern**: The stated rationale for preferring narrow-over-wide is inaccurate. The actual correct rationale — that the semantic scale has five named levels (Critical/High/Medium/Low/Trivial) with documented meaning and the handler has always enforced that — is not stated. A future reviewer reading the PR body may accept the inaccurate reasoning as precedent.
   **Severity**: nitpick

---

### What Earned Trust

The handler-level validation at `tools_store.go:59–60`, `177–178`, and `275–277` was already consistent at 0–4 before this PR. The author correctly identified the REST validator as the divergence point and fixed it there rather than introducing a third enforcement layer. The regression test `TestValidateQuickStoreInput_RejectsImportanceAbove4` covers the specific values (5, 10, 50, 100) that were previously in the gap — this is exactly the right test structure for a bug that existed in a range. The doc sweep caught the architecture.md bullet and updated it correctly. `docs/tools.md` and `docs/how-it-works.md` required no changes and were left alone appropriately.

---

### Questions Before I Sign Off

1. `handleMemoryCorrect` calls `types.ValidateImportance` (silent clamp) while `handleMemoryStore` calls the inline range check (explicit error). Was this asymmetry deliberate? If so, document why `memory_correct` should silently clamp rather than reject — the comment at `types.go:117–119` suggests the team already has a position on this.

2. The MCP tool description for `memory_correct` (`server.go:1266`) does not state the valid importance range. Do you want to add "importance 0–4" to the description string so MCP clients see it in tool discovery, consistent with how other bounded fields are documented?

3. The backward-compat break is real: any REST client calling `/quick-store` with `importance` in [5, 100] will now receive HTTP 400. Is there a CHANGELOG entry or migration note for API consumers? The PR body and commit message do not mention the breaking change.

Issue #761, opened 2026-05-20
nice-to-have: TestParseScoreLabel_ScoreErrorPropagation has no assertions


petersimmons1972:
## Summary

Found during three-round adversarial review of PR #757 (Round 2, N5 / Round 3).

`TestParseScoreLabel_ScoreErrorPropagation` in `internal/longmemeval/claude_test.go` only calls `t.Log()` — no assertions. It provides no regression protection: if `writeScoreReport` changes to handle SCORE_ERROR differently (e.g., adds a dedicated bucket), this test still passes.

The intent is to document a cross-file contract (`SCORE_ERROR` → `default` → Incorrect in `writeScoreReport`). This is better enforced with an actual assertion on the `writeScoreReport` output, or by moving the comment to a design doc and removing the empty test function.

## Suggested fix

Either:
1. Add an assertion that exercises `writeScoreReport` with a SCORE_ERROR entry and checks Incorrect count, OR
2. Convert to a comment-only design note in the relevant doc and remove the test function

## Labels
severity/nice-to-have

Issue #760, opened 2026-05-20
nice-to-have: ParseScoreLabel Pass 2 has unreachable firstLineIdx != i guard


petersimmons1972:
## Summary

Found during three-round adversarial review of PR #757 (Round 1, N2).

In `ParseScoreLabel` Pass 2, the condition `firstLineIdx >= 0 && firstLineIdx != i` contains a branch (`firstLineIdx == i`) that is unreachable in practice. Pass 1 and Pass 2 use identical comparison logic (`strings.ToUpper(strings.TrimSpace(l)) == v`). If Pass 1 failed to find a label on the first non-empty line (`firstLineIdx`), Pass 2 cannot find it there either. The guard does no harm but adds cognitive overhead.

## Suggested fix

Remove or comment the dead guard:
```go
// firstLineIdx != i is always true in Pass 2 (Pass 1 already tried firstLineIdx)
if firstLineIdx >= 0 {
    pre := ...
    ...
}
```

## Labels
severity/nice-to-have

Issue #758, opened 2026-05-20
nice-to-have: newEngineWithDims should self-register engine.Close() cleanup


petersimmons1972:
## Summary

Found during three-round adversarial review of PR #756.

The `newEngineWithDims` helper added in `internal/search/retrieval_precision_test.go` registers `backend.Close()` internally via `t.Cleanup`, but does NOT register `engine.Close()`. This is inconsistent with `newTestEngine` in `engine_test.go`, which registers both.

PR #756 works correctly because the test body adds explicit `t.Cleanup(func() { engine1024.Close() })` and `t.Cleanup(func() { engine768.Close() })`. However, future tests that call `newEngineWithDims` and forget the engine cleanup would leak goroutines silently.

## Fix

Move `t.Cleanup(func() { engine.Close() })` inside `newEngineWithDims`, matching `newTestEngine`'s pattern:

```go
func newEngineWithDims(t *testing.T, proj string, dims int) *search.SearchEngine {
    t.Helper()
    ctx := context.Background()
    backend, err := db.NewPostgresBackend(ctx, proj, testDSN(t))
    require.NoError(t, err)
    t.Cleanup(func() { backend.Close() })
    engine := search.New(ctx, backend, &fakeClient{dims: dims}, proj,
        "http://ollama:11434", "llama3.2", false, nil, 0)
    t.Cleanup(func() { engine.Close() })  // add this
    return engine
}
```

## Labels
severity/nice-to-have

PR #787, opened 2026-05-20
fix(lme): ParseScoreLabel polish (#759, #760, #761)

### LINKED ISSUE #759: nice-to-have: ParseScoreLabel Pass 2 explanation semantics when label is last line
## Summary

Found during three-round adversarial review of PR #757 (Round 1, N1).

In `ParseScoreLabel` Pass 2, when a label is found on the LAST line (old rationale-first format like `"rationale\nCORRECT"`):
- Post-label lines are empty → `explanation = ""`
- Falls to `else if pre != "" { explanation = pre }`
- Result: `explanation` becomes the pre-label rationale text

This is defensible (rationale IS the explanation in that format) but is surprising to callers who expect explanation to be post-label text. A comment or explicit test case for this boundary would prevent future confusion.

## Suggested fix

Add a test case:
```go
func TestParseScoreLabel_LabelLastLine(t *testing.T) {
    raw := "The answer matches the gold facts exactly.\nCORRECT"
    label, expl := longmemeval.ParseScoreLabel(raw)
    if label != "CORRECT" {
        t.Errorf("label = %q, want CORRECT", label)
    }
    // explanation = pre-label rationale; document this
    _ = expl
}
```

## Labels
severity/nice-to-have

petersimmons1972:
## Summary

Three `severity/nice-to-have` polish items from adversarial review of PR #757.

- **#759** — Document last-line label explanation semantics with a test. When the label is on the last line (old rationale-first format like `"rationale\nCORRECT"`), no post-label lines exist so `ParseScoreLabel` falls back to the pre-label text as the explanation. Added `TestParseScoreLabel_LabelLastLine` with assertions to document and guard this behaviour.

- **#760** — Remove unreachable `firstLineIdx != i` guard in Pass 2. Pass 1 and Pass 2 use identical comparison logic; if Pass 1 found no match at `firstLineIdx`, Pass 2 cannot match there either — the condition was always `true` when the if-block was entered. Removed the dead branch and replaced with an explanatory comment.

- **#761** — Replace the `t.Log`-only `TestParseScoreLabel_ScoreErrorPropagation` with real assertions. Three inputs that should yield `SCORE_ERROR` (truncated text, label embedded in a word, empty string) are now asserted — catches any regression where the parser incorrectly promotes an error input to a valid label and silently inflates score counts.

## Files changed

- `internal/longmemeval/claude.go` — #760 guard removed, #759 comment added (+9/−7 lines)
- `internal/longmemeval/claude_test.go` — `TestParseScoreLabel_LabelLastLine` added, `TestParseScoreLabel_ScoreErrorPropagation` replaced (+42/−11 lines)

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./internal/longmemeval/... -count=1 -race` — all pass (10/10 ParseScoreLabel tests, full suite ok)
- [x] New tests `TestParseScoreLabel_LabelLastLine` and `TestParseScoreLabel_ScoreErrorPropagation` pass

## Not included

#758 (`newEngineWithDims` cleanup) — deferred until #767 merges per coordinator brief.

Closes #759, #760, #761

### DIFF
--- a/internal/longmemeval/claude.go
+++ b/internal/longmemeval/claude.go
@@ -508,17 +508,24 @@ func ParseScoreLabel(raw string) (label, explanation string) {
 	}
 
 	// Pass 2: scan all lines for first label occurrence.
+	// Note: firstLineIdx != i is always true when the if-block below is entered —
+	// Pass 1 already tried firstLineIdx and found no match, so Pass 2 cannot
+	// match there either. The guard is removed to reduce cognitive overhead (#760).
 	for i, l := range allLines {
 		upper := strings.ToUpper(strings.TrimSpace(l))
 		for _, v := range validLabels {
 			if upper == v {
 				label = v
 				// Explanation is whatever follows on subsequent lines.
+				// When the label is on the last line (rationale-first format like
+				// "rationale\nCORRECT"), post is empty and the pre-label text
+				// becomes the explanation instead — this is intentional (#759).
 				if i+1 < len(allLines) {
 					explanation = strings.TrimSpace(strings.Join(allLines[i+1:], "\n"))
 				}
-				// Also capture any pre-label text as part of explanation if first line had content.
-				if firstLineIdx >= 0 && firstLineIdx != i {
+				// Capture any pre-label text as part of explanation.
+				// firstLineIdx != i is guaranteed (Pass 1 eliminated firstLineIdx).
+				if firstLineIdx >= 0 {
 					pre := strings.TrimSpace(strings.Join(allLines[firstLineIdx:i], "\n"))
 					if pre != "" && explanation != "" {
 						explanation = pre + "\n" + explanation
--- a/internal/longmemeval/claude_test.go
+++ b/internal/longmemeval/claude_test.go
@@ -443,23 +443,47 @@ func TestParseScoreLabel_TruncatedNoLabel(t *testing.T) {
 	}
 }
 
-// TestParseScoreLabel_ScoreErrorPropagation verifies that SCORE_ERROR returned
-// from ParseScoreLabel results in a score entry with status="error" (not
-// silently counted as PARTIALLY_CORRECT in the score report).
+// TestParseScoreLabel_LabelLastLine verifies explanation semantics when the label
+// is on the last line (rationale-first / old format like "rationale\nCORRECT").
+// When no post-label lines exist, ParseScoreLabel uses the pre-label text as the
+// explanation — this is intentional: the preamble IS the rationale. (#759)
+func TestParseScoreLabel_LabelLastLine(t *testing.T) {
+	raw := "The answer matches the gold facts exactly.\nCORRECT"
+	label, expl := longmemeval.ParseScoreLabel(raw)
+	if label != "CORRECT" {
+		t.Errorf("label = %q, want CORRECT", label)
+	}
+	// Pre-label rationale becomes the explanation in last-line-label format.
+	if expl == "" {
+		t.Errorf("explanation is empty; want pre-label rationale as explanation (#759)")
+	}
+	if !strings.Contains(expl, "matches") {
+		t.Errorf("explanation = %q; want pre-label rationale text (#759)", expl)
+	}
+}
+
+// TestParseScoreLabel_ScoreErrorPropagation verifies that ParseScoreLabel returns
+// SCORE_ERROR (not CORRECT or PARTIALLY_CORRECT) when no valid label is found.
+// Guards the pipeline contract: SCORE_ERROR hits the default/Incorrect bucket in
+// writeScoreReport (cmd/longmemeval/score.go) — never silently inflates scores. (#761)
 func TestParseScoreLabel_ScoreErrorPropagation(t *testing.T) {
-	// SCORE_ERROR should be treated as an error in writeScoreReport, not as a
-	// valid label. Verify it falls into the "default" / Incorrect bucket.
-	// This test documents the expected pipeline behaviour.
-	//
-	// In writeScoreReport (cmd/longmemeval/score.go), the switch statement:
-	//   case "CORRECT": ...
-	//   case "PARTIALLY_CORRECT": ...
-	//   default: Incorrect++
-	// SCORE_ERROR hits "default" → counted as Incorrect, which is correct
-	// behaviour (conservative: unknown = not correct).
-	//
-	// If this behaviour changes, update this comment and the switch.
-	t.Log("SCORE_ERROR falls into default/Incorrect in writeScoreReport — documented by design")
+	inputs := []string{
+		// Truncated response — context window ran out before label was emitted.
+		"The hypothesis mentions the correct city but the explanation was cut",
+		// Garbled output — label-like text embedded inside a longer word.
+		"The result is INCORRECTLY stated in the hypothesis.",
+		// Empty string — no content at all.
+		"",
+	}
+	for _, raw := range inputs {
+		label, _ := longmemeval.ParseScoreLabel(raw)
+		if label == "CORRECT" || label == "PARTIALLY_CORRECT" {
+			t.Errorf("ParseScoreLabel(%q) = %q; want SCORE_ERROR, not a valid label — would silently inflate score counts (#761)", raw, label)
+		}
+		if label != "SCORE_ERROR" {
+			t.Errorf("ParseScoreLabel(%q) = %q, want SCORE_ERROR (#761)", raw, label)
+		}
+	}
 }
 
 func TestPreferenceRecallQuery_TransformsLiteralQuestion(t *testing.T) {


petersimmons1972:
### Round 1 (Correctness) — APPROVE

**#760 — Guard reachability:** Confirmed unreachable. Pass 1 checks the first non-empty line (`firstLineIdx`) against all valid labels and `break`s if no match. Pass 2 can only match a valid label at `i` if `upper == v` — but if `i == firstLineIdx`, Pass 1 already proved that line does not match any label. The condition `firstLineIdx != i` was always `true` on entry. Removal is correct, not a regression. The replacement comment accurately describes the invariant.

**#759 — Last-line label semantics:** Traced the `"rationale\nCORRECT"` path. Post-removal, when `i+1 >= len(allLines)` the post-label explanation is empty; `firstLineIdx >= 0` is satisfied; `pre` captures the rationale text; the `else if pre != ""` branch sets `explanation = pre`. All three assertions in `TestParseScoreLabel_LabelLastLine` (`label == "CORRECT"`, `expl != ""`, `Contains(expl, "matches")`) are accurate against the post-patch code path.

**#761 — SCORE_ERROR inputs:** All three inputs correctly yield `SCORE_ERROR`.
- Truncated sentence: no line equals any valid label exactly. Returns `SCORE_ERROR`.
- `"INCORRECTLY stated..."`: `TrimSpace` + exact-equality matching means substring containment never promotes to a valid label. Returns `SCORE_ERROR`.
- Empty string: `Split("", "\n")` yields `[""]`; `firstLineIdx` stays -1; Pass 2 matches nothing. Returns `SCORE_ERROR`.

**Boundary cases (unassigned, informational):**
- Whitespace-only input: same path as empty. `SCORE_ERROR`. Safe. `[severity/nice-to-have]`
- Label in middle (`"pre\nCORRECT\npost"`): not altered by this PR; behavior unchanged and correct.

**Verdict: APPROVE — 0 blockers.** All three fixes are logically sound, the guard removal is provably safe, and the new tests assert real behavior rather than documenting intent.

petersimmons1972:

**Artifact**: PR #787 — `fix(lme): ParseScoreLabel polish (#759, #760, #761)`
**Files**: `internal/longmemeval/claude.go`, `internal/longmemeval/claude_test.go`

---

#### Trust Breakpoints

1. **Location**: `claude_test.go` — `TestParseScoreLabel_LabelLastLine` (new, line ~449) vs `TestParseScoreLabel_OldFormatRationale` (pre-existing, line ~424)
   **Observation**: Both tests use near-identical inputs (`"The hypothesis closely matches the gold answer in key facts.\nCORRECT"` vs `"The answer matches the gold facts exactly.\nCORRECT"`). `_OldFormatRationale` drops the explanation result on the floor (`label, _ := ...`). The new `_LabelLastLine` adds the explanation assertion. The old test is now a strict subset of the new one.
   **Concern**: The old test is not removed — it continues to run. This is not a bug, but it's noise. Two tests named differently, testing the same input class, one stricter than the other, with no comment linking them. A future reader will not know which is authoritative. If `_OldFormatRationale` is intentionally kept as a label-only guard independent of explanation semantics, that should be stated. If it's superseded, it should be removed.
   **Severity**: nitpick

2. **Location**: `claude.go` line 521 (pre-diff) — removal of `firstLineIdx != i` guard
   **Observation**: The PR author's claim ("Pass 1 already tried `firstLineIdx` and found no match, so Pass 2 cannot match there either") is correct for the current `validLabels` set. Pass 1 performs an exact equality check (`first == v`); Pass 2 performs the same check (`upper == v`). The normalization is identical (`strings.ToUpper(strings.TrimSpace(...))`). Given the same input, if Pass 1 returns no match at index `firstLineIdx`, Pass 2 will also return no match at that same index.
   **Concern**: The invariant holds only because Pass 1 and Pass 2 use identical normalization. If someone later changes Pass 1 to use a fuzzy or prefix match (e.g., to handle `"CORRECT."` with trailing punctuation), the invariant breaks silently — Pass 1 would match with the new logic, Pass 2 would still match at `firstLineIdx`, and the guard would have prevented double-counting the first line in the pre-label text. The replacement comment documents the current invariant, but does not document what would break it. This is a maintenance trap, not a current bug.
   **Severity**: nitpick

3. **Location**: `claude.go` — `ParseScoreLabel` and cohort functions (`ScoringPrompt`, `Score`, `ScoreResult`, `validLabels`) at lines ~441–536; `claude.go` is 833 lines
   **Observation**: The file houses HTTP client code, OAI calling conventions, batch scoring, prompt construction, and score parsing — all in one package-level file. The scoring-domain functions (`ScoringPrompt`, `Score`, `ScoreResult`, `validLabels`, `ParseScoreLabel`) are semantically cohesive and have no outward dependencies on the HTTP or generation machinery in the same file.
   **Concern**: Not a PR-introduced problem, but the file is past the point where extraction is free. `ParseScoreLabel` at 833 lines is the right time to flag it, not at 1200. Extraction to `internal/longmemeval/scoring.go` is straightforward — no circular deps, no interface changes needed — and would make the structural problem visible to future reviewers before it compounds. Flagging here per the round assignment; this is not a blocker for this PR.
   **Severity**: nitpick (structural observation, out of scope for this PR — worth a dedicated issue)

4. **Location**: `claude_test.go` — `TestParseScoreLabel_ScoreErrorPropagation` (replaced)
   **Observation**: The old test was a `t.Log`-only documentation comment with zero assertions. The new test runs three inputs through `ParseScoreLabel` and asserts `SCORE_ERROR` is returned. The original test's stated purpose was to document that `writeScoreReport` routes `SCORE_ERROR` to the `default/Incorrect` bucket in `cmd/longmemeval/score.go`. The new test no longer tests or documents that routing — it tests only the `ParseScoreLabel` return value.
   **Concern**: The original test name (`ScoreErrorPropagation`) implies end-to-end propagation through the pipeline. The new test verifies only the parser boundary. The claim "SCORE_ERROR hits `default` → counted as Incorrect" is no longer tested or asserted anywhere visible. If `writeScoreReport` is later changed to handle `SCORE_ERROR` as a special case that inflates a counter, the new test will not catch it. The behavior the old test *intended* to document is now documented only in the deleted comment. The new test is strictly better at what it tests — but it does not cover what the old test was attempting to cover, even nominally.
   **Severity**: serious — not a blocker, but the pipeline-level contract (parser → report routing) is now only in commit history, not in any live test or comment. Recommend either (a) adding a `TestScoreErrorRoutedToIncorrect` that exercises `writeScoreReport` directly, or (b) adding a comment in `TestParseScoreLabel_ScoreErrorPropagation` that explicitly acknowledges the routing contract is out of scope and references where it is tested.

---

#### What Earned Trust

- The `firstLineIdx != i` guard removal is argued correctly and the replacement comment names the invariant explicitly. That is the right level of documentation for a deleted guard.
- The new `TestParseScoreLabel_LabelLastLine` test adds real assertions where none existed and pins the pre-label-as-explanation semantic, which is a non-obvious edge case worth locking down.
- The `TestParseScoreLabel_ScoreErrorPropagation` replacement improves on a no-op test and catches the "label embedded in a word" case (`INCORRECTLY`), which is the most plausible real-world regression.
- PR description is accurate — files changed, lines changed, and test plan match the diff exactly.

---

1. `TestParseScoreLabel_OldFormatRationale` (pre-existing) and `TestParseScoreLabel_LabelLastLine` (new) test the same input class. Is `_OldFormatRationale` intentionally kept as a label-only guard, or is it superseded and should be removed?

2. The original `TestParseScoreLabel_ScoreErrorPropagation` was documenting that `writeScoreReport` routes `SCORE_ERROR` to the `default/Incorrect` bucket. Is there a test for that routing anywhere else, or is that contract now only in commit history?

3. Given `claude.go` is at 833 lines with a clearly extractable scoring-domain cluster, is there an open issue tracking `scoring.go` extraction? If not, worth filing before the file gets harder to split.

---

**Verdict**: APPROVE-WITH-NITS — 0 blockers, 1 serious finding (pipeline routing contract lost), 3 nitpicks. The serious finding is a documentation/test gap, not a logic bug, and does not gate merge per the project's severity policy. Recommend filing a tracking issue for the routing contract gap before merging.

petersimmons1972:
### Round 2 (Coverage) — APPROVE

**Coverage:** 71.1% (package `internal/longmemeval`) — well above the 55% CI gate and above the 60% new-file target.

**All 10 ParseScoreLabel tests pass** (verified on branch `fix/759-760-761-parselabel-polish`):
`Valid`, `Invalid`, `Explanation`, `Truncation`, `LabelInBody`, `MultipleLabels`, `OldFormatRationale`, `TruncatedNoLabel`, `LabelLastLine` (new), `ScoreErrorPropagation` (rewritten).

**Four rubric labels — fully covered:**
| Label | Test(s) |
|---|---|
| `CORRECT` | `Valid`, `Explanation`, `MultipleLabels`, `LabelLastLine` (new) |
| `PARTIALLY_CORRECT` | `Valid` |
| `INCORRECT` | `Valid`, `LabelInBody` |
| `SCORE_ERROR` | `Invalid`, `Truncation`, `TruncatedNoLabel`, `ScoreErrorPropagation` (rewritten) |

**`TestParseScoreLabel_LabelLastLine`** — single input, rationale-first format (`"rationale\nCORRECT"`). Sufficient: the semantics under test are a single deterministic path (no post-label lines → pre-label text becomes explanation). `TestParseScoreLabel_OldFormatRationale` already handles the label-detection path; `LabelLastLine` is correctly scoped to the explanation assignment.  `[severity/nice-to-have]` A second variant with a multi-line preamble would add marginal confidence but is not required.

**`TestParseScoreLabel_ScoreErrorPropagation`** — 3 inputs (truncated, embedded, empty). Per policy (happy + empty + boundary): empty string ✅, truncated ✅, embedded-in-word ✅. All three are boundary inputs for the SCORE_ERROR path — happy path for SCORE_ERROR is covered by `Invalid`. Rewrite converts a `t.Log`-only test into real assertions that would catch a silent inflation regression. No gaps.

**Removed `firstLineIdx != i` guard** — dead-code removal is correct; Pass 1 exhausted `firstLineIdx` before Pass 2 runs, making the condition always `true`. Explanatory comment added inline. No coverage gap introduced.

**Blockers:** 0

All findings `severity/nice-to-have` or clean. Full suite green. Coverage 71.1% ✅

petersimmons1972:

**Finding 1 — `TestParseScoreLabel_ScoreErrorPropagation` double-asserts the same condition**
For any input returning `"CORRECT"`, *both* `t.Errorf` calls fire and emit two errors for one failure. The first check is dead weight subsumed by the second. Remove the first; keep only `if label != "SCORE_ERROR"`. — `nit`

**Finding 2 — Unreachability of `firstLineIdx != i` is asserted, not shown in the diff**
The claim requires Pass 1 and Pass 2 to use strictly identical match predicates. The diff does not show Pass 1, so a reader cannot verify the invariant from this diff alone. The comment is useful but would be more durable as a reference to specific Pass 1 lines. — `severity/nice-to-have`

**Finding 3 — `#759` fix appears incomplete in the diff hunk shown**
When the label is on the last line, `i+1 < len(allLines)` is false → `explanation` stays `""`. The `if firstLineIdx >= 0` block computes `pre` but only assigns it to `explanation` when `explanation != ""` — which it is not. The visible code path does not set `explanation = pre` when the label is last. The test passes, so the assignment must exist elsewhere in the function. The comment "this is intentional (#759)" is misleading at this site. Confirm the assignment path is correct or update the comment to point at where the label-last case is actually handled. — `severity/nice-to-have`

**Finding 4 — `TestParseScoreLabel_LabelLastLine` uses `strings.Contains` as a proxy**
The `strings.Contains(expl, "matches")` assertion is brittle. An equality check against the known pre-label string would be more durable. — `nit`

**Verdict: APPROVE-WITH-NITS — 0 blockers.** Finding 3 is the highest-priority nit; worth confirming the label-last assignment path before merge.

Issue #645, opened 2026-05-15
reembed worker: silent sqlx pool hang after Postgres restart


petersimmons1972:
## Symptom

After a brief Postgres restart (~15 seconds of unavailability) during a re-embed drain on 2026-05-14, the `engram-reembed-w6800` worker silently stopped processing chunks and produced **zero log output for 10 minutes** despite:

- The container process being alive (`docker top` showed the `/engram-reembed` PID still running)
- New unleased chunks being available in the queue
- The `engram-reembed-7900xt` worker on the same Postgres recovering on its own and continuing to process

The W6800 worker only recovered when explicitly restarted (`docker restart engram-reembed-w6800`), after which the `litellm probe ok` and batch processing resumed normally.

## Timeline (UTC)

| Time | Event |
|---|---|
| 02:00:48 | Last successful `batch complete` (chunks: 2000, total_ms: 25555) |
| 02:00:30 – 02:00:32 | **6 sqlx-core warnings**: `ping on idle connection returned error: error communicating with database: expected to read 5 bytes, got 0 bytes at EOF` (Postgres restarted briefly during the burst write) |
| 02:00:48 – 02:11:xx | **Complete silence — no logs of any kind**; worker not picking up unleased chunks |
| ~02:11 | Manual `docker restart` → `litellm probe ok`, batch processing resumes |

## Reproduction context

- Triggered by a 10K-chunk re-embed campaign that pushed enough WAL/index activity to cause a brief Postgres restart
- The 7900xt worker (same image, same Postgres, different `LITELLM_URL`) self-recovered; only the W6800 worker hung
- Both workers had been running >25 minutes when the incident occurred — not a cold-start issue

## Suspected root cause

The sqlx connection pool entered a state where every checkout returned a stale/closed connection but the pool didn't tear itself down or surface an actionable error. The worker's main loop appears to silently swallow the result (treating "no available chunks" indistinguishably from "I can't talk to the database").

## Suggested fix directions

1. **Crash loud, restart fast.** If consecutive pool acquisitions fail N times within a window, the worker should panic or call `process::exit(non-zero)` so `--restart unless-stopped` brings it back fresh. Silent idle is the worst failure mode.
2. **Active pool rebuild.** On `expected to read 5 bytes, got 0 bytes at EOF` (the canonical "peer closed during query" signature), drop the pool and rebuild rather than continuing to recycle dead connections.
[...]

Issue #772, opened 2026-05-20
instinctllm/anthropic.go transport errors should wrap ErrBackendUnavailable


petersimmons1972:
client.go contract: transport errors (HTTP failure, non-200, read-body) should wrap ErrBackendUnavailable so consumers can use errors.Is to fail-soft. anthropic.go currently returns bare errors, so consolidator/detect.go's guard fails closed on infra errors with the Anthropic backend (while Olla fails open). Add ErrBackendUnavailable wrapping at the three transport error sites in anthropic.go.

PR #792, opened 2026-05-20
fix(reembed): recover from stale connection after Postgres restart (#645)

petersimmons1972:
## Diagnosis

After a Postgres restart (or a 15s network outage), all connections in the `pgxpool.Pool` become stale. When `GlobalReembedder.safeRunBatch` calls `pool.Begin()`, it gets a connection error. The worker correctly logs a warning and backs off — but the pool never drops the stale connections. On subsequent attempts the pool returns the same stale connections (until the health-check period fires), so the worker silently stops processing work.

Two compounding factors:
1. `configureSharedPool.HealthCheckPeriod` was 30s — the pool could take up to a minute to self-heal.
2. No `pool.Reset()` call on repeated errors, so the pool held stale connections across the full exponential-backoff window.

## Fix

**`internal/reembed/global_worker.go`**
- Added `consecutiveErrors atomic.Int64` counter to `GlobalReembedder`.
- In the drain loop, increment counter on error. After `consecutiveErrorThreshold` (3) consecutive failures, call `pool.Reset()` — this closes all idle connections and marks checked-out ones for close-on-return, forcing fresh dials on the next `Acquire()` call.
- Reset counter to zero on any successful batch iteration.
- Added `ConsecutiveErrors() int64` accessor for integration tests.

**`internal/db/postgres.go`**
- Tightened `configureSharedPool.HealthCheckPeriod` from 30s to 15s so the pool proactively pings and culls dead connections within one GlobalReembedder poll interval after a restart.

## Tests added

| Test | Type | DB required |
|------|------|-------------|
| `TestGlobalReembedder_ConsecutiveErrorCounterResets` | unit | no |
| `TestGlobalReembedder_HasPoolResetOnThreshold` | source guard | no |
| `TestGlobalReembedder_RecoveryIntegration` | integration | yes (TEST_DATABASE_URL) |
| `TestSharedPoolConfig_HasLifetimeSettings` | unit | no |
| `TestSharedPoolConfig_HealthCheckPeriod` | unit — regression guard for ≤15s cap | no |

The integration test calls `pool.Reset()` before starting the worker to simulate a post-restart stale pool, then asserts the worker processes a NULL-embedded chunk within 25 seconds. It skips if `TEST_DATABASE_URL` is unset.

## Files changed

- `internal/reembed/global_worker.go` (+40 lines)
- `internal/reembed/global_worker_recovery_test.go` (new, +155 lines)
- `internal/db/postgres.go` (+3 lines)
- `internal/db/postgres_pool_config_test.go` (+46 lines)
- `internal/db/postgres_shared_pool_test.go` (+7 lines)

### DIFF
--- a/internal/instinctllm/anthropic.go
+++ b/internal/instinctllm/anthropic.go
@@ -93,17 +93,20 @@ func (c *anthropicClient) Complete(ctx context.Context, systemPrompt, userPrompt
 
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return "", fmt.Errorf("llm/anthropic: HTTP: %w", err)
+		slog.Warn("llm/anthropic: transport error", "err", err)
+		return "", fmt.Errorf("llm/anthropic: transport: %w", ErrBackendUnavailable)
 	}
 	defer func() { _ = resp.Body.Close() }()
 
 	if resp.StatusCode != http.StatusOK {
-		return "", fmt.Errorf("llm/anthropic: non-200 status: %d", resp.StatusCode)
+		slog.Warn("llm/anthropic: non-200 status", "status", resp.StatusCode)
+		return "", fmt.Errorf("llm/anthropic: status %d: %w", resp.StatusCode, ErrBackendUnavailable)
 	}
 
 	raw, err := io.ReadAll(resp.Body)
 	if err != nil {
-		return "", fmt.Errorf("llm/anthropic: read body: %w", err)
+		slog.Warn("llm/anthropic: read body", "err", err)
+		return "", fmt.Errorf("llm/anthropic: read body: %w", ErrBackendUnavailable)
 	}
 
 	var apiResp struct {
--- a/internal/instinctllm/anthropic_test.go
+++ b/internal/instinctllm/anthropic_test.go
@@ -179,3 +179,77 @@ func TestAnthropicEmptyPrompts(t *testing.T) {
 		t.Errorf("empty user prompt: got %q, want []", got)
 	}
 }
+
+// ---------------------------------------------------------------------------
+// #772 — transport errors must wrap ErrBackendUnavailable
+// ---------------------------------------------------------------------------
+
+// TestAnthropicTransportError_WrapsBackendUnavailable: server is closed before
+// the request lands (connection refused). The error must wrap ErrBackendUnavailable
+// so errors.Is(err, ErrBackendUnavailable) returns true.
+func TestAnthropicTransportError_WrapsBackendUnavailable(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	// Close immediately so subsequent request gets connection refused.
+	srv.Close()
+
+	c := newAnthropicClient(t, srv.URL+"/v1/messages")
+	_, err := c.Complete(context.Background(), "system", "user")
+	if err == nil {
+		t.Fatal("Complete() should return error when server is unavailable")
+	}
+	if !errors.Is(err, instinctllm.ErrBackendUnavailable) {
+		t.Errorf("transport error must wrap ErrBackendUnavailable; got: %v", err)
+	}
+}
+
+// TestAnthropicNon200_WrapsBackendUnavailable: server returns 503; error must
+// wrap ErrBackendUnavailable so callers can use errors.Is to skip gracefully.
+func TestAnthropicNon200_WrapsBackendUnavailable(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusServiceUnavailable)
+	}))
+	defer srv.Close()
+
+	c := newAnthropicClient(t, srv.URL+"/v1/messages")
+	_, err := c.Complete(context.Background(), "system", "user")
+	if err == nil {
+		t.Fatal("Complete() should return error on 503")
+	}
+	if !errors.Is(err, instinctllm.ErrBackendUnavailable) {
+		t.Errorf("non-200 status must wrap ErrBackendUnavailable; got: %v", err)
+	}
+}
+
+// TestAnthropicReadBodyError_WrapsBackendUnavailable: server closes the
+// connection mid-response (hijack), so ReadAll fails. The error must wrap
+// ErrBackendUnavailable.
+func TestAnthropicReadBodyError_WrapsBackendUnavailable(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Send a valid 200 header but then hijack and close so body read fails.
+		hj, ok := w.(http.Hijacker)
+		if !ok {
+			t.Error("ResponseWriter does not implement Hijacker")
+			return
+		}
+		conn, _, err := hj.Hijack()
+		if err != nil {
+			t.Errorf("Hijack: %v", err)
+			return
+		}
+		// Write minimal HTTP/1.1 200 headers then close mid-body.
+		_, _ = conn.Write([]byte("HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nContent-Length: 100\r\n\r\n{"))
+		conn.Close()
+	}))
+	defer srv.Close()
+
+	c := newAnthropicClient(t, srv.URL+"/v1/messages")
+	_, err := c.Complete(context.Background(), "system", "user")
+	if err == nil {
+		t.Fatal("Complete() should return error on truncated body")
+	}
+	if !errors.Is(err, instinctllm.ErrBackendUnavailable) {
+		t.Errorf("read-body failure must wrap ErrBackendUnavailable; got: %v", err)
+	}
+}


petersimmons1972:

**Scope note:** The diff touches `internal/instinctllm/anthropic.go` (+3 changed error paths) and `internal/instinctllm/anthropic_test.go` (+3 new tests). The PR description references `global_worker.go` and `postgres.go` — those files are not in the diff. The description appears to be from a different feature branch and was pasted into this PR by mistake. This is an administrative issue, not a logic defect; the code under review is internally consistent and sound.

---

#### Findings

**[NIT] Read-body error loses original cause**

In the proposed change, the `io.ReadAll` failure path becomes:
```go
return "", fmt.Errorf("llm/anthropic: read body: %w", ErrBackendUnavailable)
```
The original transport error (`err`) is logged via `slog.Warn` but is not preserved in the returned error chain. The transport and non-200 paths do the same. This is consistent with the `olla.go` pattern, so it is not a regression, but it means `errors.Unwrap` on the returned error yields only `ErrBackendUnavailable` — the root OS/network error is dropped from the chain. Future debugging requires correlating the structured log entry. Acceptable as-is; worth noting.

**[NIT] Context-cancellation path now wraps ErrBackendUnavailable**

After this change, a context-cancelled or context-deadline-exceeded call to `http.DefaultClient.Do` will return a transport error, which the new code wraps as `ErrBackendUnavailable`. The existing `TestAnthropicContextCancellation` test only asserts `err != nil` — it does not assert `!errors.Is(err, ErrBackendUnavailable)`. The caller in `detect.go` treats `ErrBackendUnavailable` as "skip batch silently" — so a context cancellation (e.g. shutdown signal) will now cause the batch to be silently skipped rather than propagating cancellation up the call stack. This may be intentional (matches `olla.go` which does the same), but it is a semantic change worth confirming. No test currently verifies this caller-visible behavior for context cancellation specifically on the Anthropic path.

**[NIT] `TestAnthropicErrorPropagation` (existing) tests 500, not `ErrBackendUnavailable` wrapping**

The existing test at line 55 verifies only `err != nil` on a 500 response. After this PR, a 500 will now return an error wrapping `ErrBackendUnavailable`. The test does not assert this, so the new sentinel contract for 5xx responses goes untested in the existing test. The three new tests cover 503, connection refused, and body truncation — but a plain 500 (internal server error) is a different semantic case and is not verified to wrap the sentinel.

**[PASS] Error-chain correctness for transport, non-200, and read-body paths** — all three now correctly wrap `ErrBackendUnavailable` so `errors.Is` returns true. Consistent with `olla.go` pattern.

**[PASS] Empty-content and parse-failure paths correctly do NOT wrap `ErrBackendUnavailable`** — semantically correct: these are model-behaviour conditions, not infrastructure failures. Pre-existing tests `TestAnthropicEmptyContentNotSentinel` and `TestAnthropicParseFailureIsNotSentinel` provide regression coverage.

**[PASS] `TestAnthropicReadBodyError_WrapsBackendUnavailable` hijack technique** — valid mechanism for simulating mid-response body truncation. The test is not brittle; it tests the actual wrapping contract, not source-level presence of any call.

---

#### Verdict

No blockers. Three nits, all severity/nice-to-have. The primary concern worth a quick author confirmation: is silent skip-on-cancel (via `ErrBackendUnavailable` wrapping of context errors) intentional for the Anthropic path, matching the existing `olla.go` behavior?

Also: please update the PR description to reflect the actual files changed — the current description describes a completely different fix (pgxpool stale-connection recovery) and will confuse future reviewers reading git history.

petersimmons1972:

**Artifact**: PR #792 — `fix(reembed): recover from stale connection after Postgres restart (#645)`
**Files reviewed**: `internal/reembed/global_worker.go`, `internal/reembed/global_worker_recovery_test.go`, `internal/db/postgres.go`, `internal/db/postgres_pool_config_test.go`, `internal/db/postgres_shared_pool_test.go`

---

### Trust Breakpoints

**1. Recovery placement: worker calls Reset() instead of pool-level encapsulation**
**Location**: `internal/reembed/global_worker.go:148`
**Observation**: `g.pool.Reset()` is called directly inside the GlobalReembedder drain loop. The per-project `Worker` (`internal/reembed/worker.go`) has no equivalent recovery; it has `safeRunBatch` panic recovery but no stale-connection Reset path.
**Concern**: The fix works for GlobalReembedder but does not protect any per-project Worker against the same failure mode. A Postgres restart will still silently stall per-project Workers. The correct architectural home is either `db.NewSharedPool` (wrapping the pool in a type that auto-resets on repeated acquire failures) or a shared recovery helper both Workers call. Scattered worker-level Reset calls are a maintenance smell — when a third consumer is added this will be missed again. This is a design liability, not a correctness bug in this PR.
**Severity**: serious

---

**2. Double-Reset race: threshold check and Reset() are not atomic**
**Location**: `internal/reembed/global_worker.go:134–149`
**Observation**: `consecutive := g.consecutiveErrors.Add(1)` is atomic. But `if consecutive >= consecutiveErrorThreshold { g.pool.Reset() }` is a separate operation. The GlobalReembedder runs a single drain goroutine, so two goroutines cannot race here in the current implementation. However, `ConsecutiveErrors()` is exported and the comment "Exposed for integration tests" implies external observers. The goroutine structure is implicit — nothing prevents a future caller from calling `safeRunBatch` directly in a second goroutine.
**Concern**: The atomicity guarantee is real today but fragile by design. The exported `ConsecutiveErrors()` accessor signals that the counter is part of the public API, which invites exactly the kind of misuse that would create a double-Reset race. `pool.Reset()` on a live pool during active `Acquire()` calls is documented safe by pgxpool, but a double call within milliseconds wastes connection establishment work under high load. Worth a comment that this is safe only because the drain loop runs in a single goroutine.
**Severity**: nitpick

---

**3. HealthCheckPeriod 15s is now globally coupled to GlobalReembedder's poll interval**
**Location**: `internal/db/postgres.go:65`, comment: "15s health-check so the pool detects dead connections within one GlobalReembedder poll interval"
**Observation**: The pool configuration comment encodes a dependency on `GlobalReembedder`'s interval. If someone changes the GlobalReembedder poll interval without updating the pool config (or vice versa), the comment becomes silently wrong and recovery degrades.
**Concern**: This cross-module coupling is invisible at compile time. A developer reading `postgres.go` has no link to `global_worker.go`'s interval constant, and vice versa. The "15s" rationale only makes sense if you know the reembedder polls at ~15s. The test `TestSharedPoolConfig_HealthCheckPeriod` hard-gates ≤15s, but it does not enforce any relationship to the worker interval — if the worker interval changes to 5s, the pool config becomes under-tuned with no failing test.
**Severity**: serious

---

**4. Source-level presence test is an antipattern**
**Location**: `internal/reembed/global_worker_recovery_test.go:51–64` (`TestGlobalReembedder_HasPoolResetOnThreshold`)
**Observation**: This test opens `global_worker.go` with `os.ReadFile` and checks for `strings.Contains(text, "pool.Reset()")`. It tests that a string appears in source, not that the function executes.
**Concern**: This is a smell the author acknowledged ("source guard"). The test would pass if `pool.Reset()` was commented out with `// pool.Reset()`, if it was moved into a helper called `resetPool()`, or if `consecutiveErrorThreshold` was renamed. It provides false confidence: a reviewer who sees this test pass assumes recovery is in place; a simple refactor breaks the guarantee silently. This pattern has no precedent in this repo — every other test exercises behavior. The `TestGlobalReembedder_RecoveryIntegration` test is the real guard; this source-scan test adds noise without safety. It should be deleted or replaced with a unit test using an injectable pool interface.
**Severity**: serious

---

**5. Integration test polls with `time.Sleep(200ms)` and a 25s wall-clock deadline**
**Location**: `internal/reembed/global_worker_recovery_test.go:141–154`
**Observation**: The test loop polls via `time.Sleep(200*time.Millisecond)` up to 25 seconds. The test does not use `t.Deadline()` or `context` propagation from the outer `context.WithTimeout`.
**Concern**: The outer context is 30s; the inner deadline is 25s hardcoded via `time.Now()`. If the test machine is slow (CI under load), the 30s context can expire before the 25s deadline is reached, causing a confusing "context canceled" error instead of the informative `t.Errorf` message. The poll loop should select on the outer ctx.Done() and call `t.Fatalf("context expired: %v", ctx.Err())` to preserve diagnostic quality.
**Severity**: nitpick

---

### What Earned Trust

- `consecutiveErrors atomic.Int64` is the right primitive for a single writer — `Add` and `Store` are sufficient; no mutex needed given the single-goroutine drain loop. The author chose correctly.
- `pool.Reset()` placement inside the drain loop (not inside `safeRunBatch`) is deliberate and correct: the counter must accumulate across retries within one tick, and resetting within the batch would conflate per-chunk errors with per-pool errors.
- `TestSharedPoolConfig_HealthCheckPeriod` using `≤ maxAllowed` (upper bound) rather than `== 15s` is the right shape for a regression guard.
- The `configureSharedPool` / `configurePool` split already existed; the PR correctly places the HealthCheckPeriod change in `configureSharedPool` rather than the CLI-tool pool. No blast radius on unrelated consumers.
- `ConsecutiveErrors()` accessor is narrow and read-only; it does not expose the counter for external mutation.

---

### Questions Before I Sign Off

1. **Per-project Worker**: `internal/reembed/worker.go` has no stale-connection recovery. Is this intentional (per-project workers are expected to restart on error)? Or is this a known gap that should be tracked as a follow-up issue?

2. **Pool-level abstraction**: Was a pool-level wrapper (e.g., a `ResilientPool` type that auto-resets on consecutive acquire failures) considered? What ruled it out?

3. **Source guard test**: Is `TestGlobalReembedder_HasPoolResetOnThreshold` intended to stay permanently, or is it a placeholder until an injectable pool interface is added? If it's permanent, the team should acknowledge this pattern explicitly — it has no other precedent here.

4. **HealthCheckPeriod coupling**: Is there a mechanism (constant reference, build tag, test assertion) planned to keep the 15s pool config in sync with the GlobalReembedder poll interval? Right now they are independent magic numbers that happen to match.

petersimmons1972:
### Round 2 (Coverage) — BLOCK

**Scope mismatch is a blocker.** The PR title, body, and commit graph are inconsistent.

---

#### Critical finding

| Severity | Finding |
|----------|---------|
| `severity/blocker` | **PR body describes changes that are not in the diff.** The body documents `internal/reembed/global_worker.go` (+40 lines, `consecutiveErrors` counter, `pool.Reset()` on threshold), `internal/db/postgres.go` (HealthCheckPeriod 30s→15s), `internal/reembed/global_worker_recovery_test.go` (+155 lines), and two `postgres_*_test.go` files. None of these files appear in the actual diff. The committed change is exclusively `internal/instinctllm/anthropic.go` — a separate fix wrapping `ErrBackendUnavailable` on transport errors. The fix described in the body (stale-connection recovery for `GlobalReembedder`) was either never committed or was committed to the wrong branch. Reviewing "coverage on global_worker.go" is impossible when that file has no PR changes. |

---

#### What IS in the diff (instinctllm transport-error fix)

Three files changed: `internal/instinctllm/anthropic.go` (+9/-3), `internal/instinctllm/anthropic_test.go` (+74), `results/README.md` (doc only).

**Coverage — `internal/instinctllm/` package**

| Metric | Value | Gate |
|--------|-------|------|
| Package statement coverage | **84.8%** | ≥55% CI gate, ≥60% target |
| `anthropic.go:Complete` function | **94.7%** | — |
| `anthropic.go:NewAnthropicClient` | 100.0% | — |

The 5.3% gap in `Complete` is two effectively-untestable branches: `json.Marshal` failure and `http.NewRequestWithContext` failure. These are not a concern.

**Tests added**

All 3 new tests pass and exercise the correct error paths:

- `TestAnthropicTransportError_WrapsBackendUnavailable` — connection refused → `errors.Is(err, ErrBackendUnavailable)` ✅
- `TestAnthropicNon200_WrapsBackendUnavailable` — 503 status → wraps sentinel ✅
- `TestAnthropicReadBodyError_WrapsBackendUnavailable` — mid-body hijack close → wraps sentinel ✅

The instinctllm change itself is solid: correct sentinel wrapping, tests are meaningful (not brittle string guards), no counter-reset or source-level presence tests needed here.

**No integration test, no `pool.Reset()` source guard, no counter-reset unit tests** — these belong to the reembed fix described in the body, which is absent from the diff.

---

#### Resolution path

Either (a) push the missing reembed/postgres changes onto this branch so the body matches the diff, or (b) split: merge the instinctllm fix separately (rename the PR to "fix(instinctllm): wrap transport errors with ErrBackendUnavailable") and open a new PR for the GlobalReembedder stale-connection recovery. Option (b) is cleaner given these are independent fixes.

The instinctllm portion would pass Round 2 on its own — zero coverage blockers, all tests green, 84.8% package coverage.

---

**Verdict: BLOCK** — 1 blocker (`severity/blocker`: PR body describes absent changes). 0 blockers on the instinctllm portion in isolation. Coverage 84.8%.

petersimmons1972:
Closing: branch contained a duplicate of #772's anthropic.go fix (also on PR #797), not the #645 reembed fix described in the title/body. Re-opening the #645 fix on a fresh branch.

Issue #729, opened 2026-05-18
arch: eliminate multiple non-authoritative sources for the same config values (single source of truth)


petersimmons1972:
## Severity: Blocking

Configuration values defined in multiple places create silent failure modes. The `ENGRAM_HOST` issue (#728) is the canonical example: the default in `cmd/engram/main.go` was `127.0.0.1` but nothing in `docker-compose.yml` overrode it, causing Engram to be unreachable via docker-proxy for an unknown period. Docker's healthcheck was a false positive the entire time.

## Pattern: Three-Layer Definition

Current pattern for most values:
```
code default → env var → compose environment block → .env file
```

When any layer has a different value, the one closest to the binary wins. If a middle layer is missing, the code default silently takes over.

## Known Instances

| Value | Code default | Compose | .env | Risk |
|-------|-------------|---------|------|------|
| ENGRAM_HOST | 127.0.0.1 | 0.0.0.0 (added in #728) | not set | Root cause of HTTP unreachable |
| ENGRAM_PORT | 8788 | \${ENGRAM_PORT:-8788} | ENGRAM_PORT=8788 | Three-way tie — no conflict yet |
| LITELLM_URL | n/a | \${LITELLM_URL:-http://olla:40114} | http://192.168.0.138:30411 | Compose default would fail if .env missing |
| ENGRAM_EMBED_URL | n/a | \${ENGRAM_EMBED_URL:-http://example.invalid:8003} | set | Placeholder would silently poison embed |

## Required Audit

1. Find all compose defaults: grep `:-` in docker-compose*.yml
2. Find all code-level defaults: grep `envInt\|envOr\|os.Getenv` in cmd/
3. Cross-reference with .env.example
4. Check related bugs #544, #666, #668, #687 for same pattern

## Fix Approach

- For required values: remove compose defaults; fail fast (`\${VAR:?VAR must be set}`)
- For values with safe defaults: document chain explicitly; log which layer resolved each value at startup
- Consider: single config.go as the only place defaults are defined

## Note

Do not fix tonight. Benchmark must resume first.

Issue #728, opened 2026-05-18
fix: Engram MCP server binds to 127.0.0.1 inside container — HTTP unreachable via docker-proxy


petersimmons1972:
## Root Cause

`ENGRAM_HOST` defaults to `127.0.0.1` (per `cmd/engram/main.go` + #666 comment). Compose did not override it.

The port mapping `127.0.0.1:8788:8788` forwards host loopback → container eth0 (`172.26.x`). The Go server binds only to the container's loopback (`127.0.0.1:8788`), not eth0. docker-proxy sends packets to eth0:8788 → nothing listening → RST.

Observed:
- Docker shows `healthy` (false positive — healthcheck CLI doesn't make an HTTP call)
- `curl http://127.0.0.1:8788/health` from host: **connection reset by peer**
- App log: `"engram MCP server starting","addr":"127.0.0.1:8788"`
- MCP client disconnects immediately on SSE connect

## Fix Applied

Added `ENGRAM_HOST=0.0.0.0` to `docker-compose.yml` environment block on `feat/instinct-canonical` (commit `ad6c1bc`).

Port mapping still restricts host access to `127.0.0.1` — no security surface increase.

## Secondary Issues Identified

1. **Reembed workers crashlooping** — `connect to postgres` error every minute. Likely POSTGRES_PASSWORD not propagated to reembed env. Investigate: `docker compose exec engram-reembed-7900xt /engram-reembed --probe`
2. **Summarize worker 404 spam** — `llama3.2:3b` not routed in olla. Set `ENGRAM_CLAUDE_SUMMARIZE=true` or add olla route.

## Status

Fix committed, rebuild + container restart pending.

petersimmons1972:
Fixed in commit ad6c1bc — added `ENGRAM_HOST=0.0.0.0` to the engram-go service in docker-compose.yml. MCP server now binds all interfaces; docker-proxy forwards correctly. Verified: Engram MCP reachable at http://localhost:8788 from host.

Issue #666, opened 2026-05-17
ENGRAM_RATE_LIMIT_DISABLE=true + 0.0.0.0 bind allows LAN bearer-token brute-force


petersimmons1972:
**Reporter**: Security Reviewer (High)

**Location**: `.env:51` (`ENGRAM_RATE_LIMIT_DISABLE=true`); `cmd/engram/main.go:102` (`ENGRAM_HOST` defaults `0.0.0.0`); `docker-compose.yml` `ports: "${ENGRAM_PORT:-8788}:8788"` (no `127.0.0.1:` prefix); corroborated by `internal/mcp/server.go:780-784`

Server logs `HTTP rate limiter disabled via ENGRAM_RATE_LIMIT_DISABLE (#387)` at startup. With `0.0.0.0` bind and host-level port publish (no loopback prefix), any LAN host can hammer the MCP endpoint or bruteforce the 64-char bearer token at full line rate.

SECURITY.md says "Loopback binding by default" — compose file does not enforce this.

**Required**:
1. Either remove `ENGRAM_RATE_LIMIT_DISABLE` from `.env` OR document the threat model accepting it
2. Default compose port publish to `127.0.0.1:8788:8788`
3. Reconcile SECURITY.md vs. compose default behavior

---
**Source**: QA persona sweep 2026-05-17 (commit f0ce4e0)

PR #794, opened 2026-05-20
refactor(config): consolidate ENGRAM_* config sources (#729)

petersimmons1972:
## Summary

- Introduces `internal/config` as the **single source of truth** for `ENGRAM_HOST` and `ENGRAM_PORT` defaults. No more inline string/int literals scattered across binary entrypoints.
- Fixes a real bug in `cmd/starter` `runHealth()`: it hardcoded port `"8788"` and never read `ENGRAM_PORT`, so a non-default port would cause the Docker `HEALTHCHECK` to probe the wrong address — silently returning false-positives.
- `cmd/engram/main.go` flag defaults now reference `config.DefaultHost` / `config.DefaultPort` / `config.EnvKey*` constants.

## Config sources before this PR

| Source | ENGRAM_HOST | ENGRAM_PORT |
|--------|-------------|-------------|
| `cmd/engram/main.go` inline literal | `"127.0.0.1"` | `8788` |
| `cmd/starter/main.go` runHealth() | *(hardcoded `"8788"`, ENGRAM_PORT not read)* | `"8788"` (hardcoded string) |
| `docker-compose.yml` explicit override | `0.0.0.0` | `${ENGRAM_PORT:-8788}` |

## Config sources after this PR

| Source | ENGRAM_HOST | ENGRAM_PORT |
|--------|-------------|-------------|
| `internal/config/config.go` (**canonical**) | `"127.0.0.1"` | `8788` |
| `cmd/engram/main.go` | references `config.DefaultHost` | references `config.DefaultPort` |
| `cmd/starter/main.go` runHealth() | `127.0.0.1` (unchanged — correct for loopback probe) | reads `ENGRAM_PORT` via `config.PortOrDefault()` |
| `docker-compose.yml` | `ENGRAM_HOST=0.0.0.0` (intentional container override) | `${ENGRAM_PORT:-8788}` (passthrough) |

## Env vars consolidated

- `ENGRAM_HOST` — constant `config.EnvKeyHost`, default `config.DefaultHost = "127.0.0.1"`
- `ENGRAM_PORT` — constant `config.EnvKeyPort`, default `config.DefaultPort = 8788`

## Files changed

| File | Change |
|------|--------|
| `internal/config/config.go` | New — canonical defaults + typed accessors |
| `internal/config/config_test.go` | New — 12 tests, 100% function coverage |
| `cmd/engram/main.go` | Replace inline literals with `config.*` constants (+3/-1) |
| `cmd/starter/main.go` | Fix hardcoded port in `runHealth()` (+6/-2) |

## Follow-up (out of scope)

The remaining scattered `os.Getenv` calls in `internal/mcp/server.go` and `internal/chunk/chunker.go` can be migrated to `envconf.*` helpers as a separate PR once the scope of this one is confirmed clean.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./internal/config/... -count=1 -race` — 12/12 pass
- [x] `go test ./... -count=1 -race -short` — all pass
- [x] `docker compose config --quiet` — YAML parses cleanly

### DIFF
--- a/cmd/audit/main.go
+++ b/cmd/audit/main.go
@@ -25,8 +25,8 @@ func main() {
 func run() int {
 	llmBackend := flag.String("llm-backend", envOr("LLM_BACKEND", "anthropic"), "LLM backend: anthropic or olla")
 	timeout := flag.Duration("timeout", defaultTimeout, "Per-inference timeout")
-	engramBase := flag.String("engram", "", "Engram base URL (overrides config)")
-	engramToken := flag.String("token", "", "Engram Bearer token (overrides config)")
+	engramBase := flag.String("engram", "", "Engram base URL (overrides ENGRAM_URL env var and ~/.claude/mcp_servers.json)")
+	engramToken := flag.String("token", "", "Engram Bearer token (overrides ENGRAM_TOKEN env var and ~/.claude/mcp_servers.json)")
 	flag.Parse()
 
 	// Resolve Engram coordinates.
--- a/cmd/audit/report.go
+++ b/cmd/audit/report.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"os"
 	"path/filepath"
 	"strings"
@@ -72,18 +73,41 @@ func envOr(key, fallback string) string {
 	return fallback
 }
 
+// errNoEngram is returned when no Engram credentials can be resolved from
+// flags, environment variables, or the developer config file.
+var errNoEngram = errors.New(
+	"Engram credentials not found: set ENGRAM_URL and ENGRAM_TOKEN env vars, " +
+		"or pass -engram and -token flags",
+)
+
 // resolveEngram returns the Engram base URL and bearer token to use.
-// If baseFlag and tokenFlag are both non-empty they are used as-is.
-// Otherwise the values are read from ~/.claude/mcp_servers.json.
-// Ported verbatim from instinct-python/cmd/audit/main.go:262-284.
+//
+// Resolution order (first fully-populated pair wins):
+//  1. -engram / -token CLI flags (baseFlag, tokenFlag)
+//  2. ENGRAM_URL / ENGRAM_TOKEN environment variables
+//  3. ~/.claude/mcp_servers.json — developer-only fallback, attempted only
+//     when the file exists and no env vars are set
+//
+// If none of the above provide both values, errNoEngram is returned.
 func resolveEngram(baseFlag, tokenFlag string) (string, string, error) {
+	// 1. Flags take priority.
 	if baseFlag != "" && tokenFlag != "" {
 		return baseFlag, tokenFlag, nil
 	}
+
+	// 2. Environment variables.
+	envURL := os.Getenv("ENGRAM_URL")
+	envToken := os.Getenv("ENGRAM_TOKEN")
+	if envURL != "" && envToken != "" {
+		return envURL, envToken, nil
+	}
+
+	// 3. Developer config file — silent fallback when the file exists.
 	cfgPath := filepath.Join(os.Getenv("HOME"), ".claude", "mcp_servers.json")
 	data, err := os.ReadFile(cfgPath)
 	if err != nil {
-		return "", "", err
+		// File absent or unreadable — not a developer machine.
+		return "", "", errNoEngram
 	}
 	var cfg struct {
 		McpServers map[string]struct {
@@ -97,5 +121,8 @@ func resolveEngram(baseFlag, tokenFlag string) (string, string, error) {
 	e := cfg.McpServers["engram"]
 	sseURL := strings.TrimSuffix(strings.TrimRight(e.URL, "/"), "/sse")
 	auth := strings.TrimPrefix(e.Headers["Authorization"], "Bearer ")
+	if sseURL == "" || auth == "" {
+		return "", "", errNoEngram
+	}
 	return sseURL, auth, nil
 }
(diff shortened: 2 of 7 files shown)


petersimmons1972:
Branch cleanup: removed two out-of-scope commits that were accidentally landing here — the #645 reembed fix (now on PR #798, branch `fix/645-reembed-worker-hang-v2`) and a duplicate of the #768 importance-bounds fix (already merged via PR #786). PR is now scoped to #729 config SSoT only. Force-pushed cleanly — single commit `98f98a0`.

petersimmons1972:
### Consolidated 4-Round Review — APPROVE (no blockers)

**PR:** refactor(config): consolidate ENGRAM_* config sources (#729)
**Scope:** New `internal/config` package + wiring into two binary entrypoints. Real bug fix in `runHealth()`.

---

#### Round 1 — Correctness

**Bug fix is correct.** `runHealth()` hardcoded `"8788"` and never read `ENGRAM_PORT`. The fix replaces this with `config.PortOrDefault()` which reads the env var and falls back to the canonical default. The `fmt.Sprintf` switch from string concatenation is a minor style improvement with no semantic change.

**`Port()` error handling:** The function returns `(0, error)` when ENGRAM_PORT is set but malformed. The caller `PortOrDefault()` silently falls back to `DefaultPort` on any error. This is fine for `runHealth()` — a silent fallback is correct here (the health binary should not crash on a bad env var; it should probe the default port and let the operator notice via a failed health check). For `cmd/engram/main.go`, the port is read via `envInt(config.EnvKeyPort, config.DefaultPort)` which is the existing helper, not `config.Port()` — so `config.Port()`'s error path is not exercised in the main binary's flag parsing. This is consistent but worth noting: a malformed ENGRAM_PORT in the server binary falls through to `flag.Int` default handling, not `config.Port()`'s error. No blocker.

**`Host()` validation:** No validation on the host string (empty after trim, invalid IP, etc.). For a config accessor this is acceptable — the OS will reject a bad bind address at `net.Listen` time with a clear error. No blocker.

**Nil/zero boundary:** `Port()` rejects `n <= 0`, so port 0 (OS-assigned) is explicitly rejected. This is a deliberate policy choice and is documented. Acceptable.

**Verdict:** No correctness blockers.

---

#### Round 2 — Coverage

`internal/config/config_test.go` claims 12 tests — the diff shows exactly 12 test functions. Coverage breakdown:

- `DefaultHost_Value`, `Host_Unset`, `Host_Set` — 3/3 for `Host()`
- `DefaultPort_Value`, `Port_Unset`, `Port_Valid`, `Port_Malformed`, `Port_Zero`, `Port_Negative` — 6/6 for `Port()`
- `PortOrDefault_Unset`, `PortOrDefault_Valid`, `PortOrDefault_Malformed` — 3/3 for `PortOrDefault()`

Missing cases worth noting (nice-to-have, not blockers):
- `ENGRAM_PORT` set to whitespace-only (e.g. `"   "`) — `strings.TrimSpace` before `strconv.Atoi` handles this, but there is no test asserting the behavior.
- `ENGRAM_HOST` set to whitespace-only — `Host()` would return the whitespace string since it only checks `v != ""`. Probably harmless in practice.

Per CLAUDE.md, the PR body claims 100% function coverage on the new package. The test file covers all three exported functions plus both constants. The bug fix in `runHealth()` does not have a dedicated integration test, but the unit tests for `PortOrDefault()` cover the code path. The PR body's test plan lists `go test ./internal/config/... -count=1 -race` passing — the tests exist and are deterministic.

**Verdict:** Coverage adequate. Two nice-to-have whitespace edge cases.

---

#### Round 3 — Structural

The package is the right size: two constants, two env-key constants, three accessors, one file. It does not try to own all of `ENGRAM_*` at once — the PR body explicitly calls out the remaining `os.Getenv` calls in `mcp/server.go` and `chunk/chunker.go` as out-of-scope follow-up. This is correct scope discipline.

`EnvKeyHost`/`EnvKeyPort` as exported constants (rather than unexported) is the right call — callers in `cmd/` need them for `envOr`/`envInt` helpers.

The package doc comment is clear and contains a usage example. The `DefaultHost` comment explaining the Docker container bind-address decision (reference to #666, #728) is valuable institutional knowledge in the right place.

The diff to `cmd/engram/main.go` is minimal (+3/-1 in the flag registration section). The `cmd/starter/main.go` change is clean: `runHealth()` now has a comment explaining why `PortOrDefault()` is used here and a reference to #729. The `fmt.Sprintf` change is a bonus cleanup.

**Verdict:** No structural blockers. Package scope and naming are appropriate.

---

#### Round 4 — Zero-Context Fresh Eyes

Reading only the diff: a new `internal/config` package defines canonical defaults and env-var names for host and port. Two binary entrypoints are updated to use the constants. One function (`runHealth`) is fixed to read the env var instead of using a hardcoded string. Tests cover all exported symbols. 

Questions a zero-context reviewer would ask:
1. *Is `PortOrDefault()` the right API for `runHealth()`?* Yes — a health binary should not crash on a misconfigured env var.
2. *Does any deployment file need to be updated?* The PR body table shows `docker-compose.yml` uses `${ENGRAM_PORT:-8788}` which is a passthrough, not a hardcoded default — it remains correct.
3. *Could this break existing deployments?* No. Defaults are identical to the previous inline literals. The only behavioral change is `runHealth()` now reading `ENGRAM_PORT`, which is strictly an improvement.

**Verdict:** No blockers from fresh-eyes pass.

---

#### Summary

| Round | Result |
|-------|--------|
| Correctness | ✅ No blockers — `runHealth()` bug fix is correct |
| Coverage | ✅ 12 tests, 100% function coverage on new package |
| Structural | ✅ Correct scope, clean naming |
| Zero-context | ✅ Safe for existing deployments |

**APPROVE.** No `severity/blocker` findings. Nice-to-have: add tests for whitespace-only `ENGRAM_PORT`/`ENGRAM_HOST` env values.

PR #798, opened 2026-05-20
fix(reembed): recover from stale connection after Postgres restart (#645)

### LINKED ISSUE #645: reembed worker: silent sqlx pool hang after Postgres restart

petersimmons1972:
## Problem

After a Postgres restart (or any outage ≥15 s), all pool connections become stale. `pool.Begin()` fails, `safeRunBatch` returns an error, and the worker backs off — but without calling `pool.Reset()` the pool never drops stale connections. The worker silently stops processing reembed work until the process is restarted.

Closes #645

## Fix

- **Atomic consecutive-error counter**: increments on each `safeRunBatch` failure, resets to 0 on success.
- **`pool.Reset()` on threshold**: after `consecutiveErrorThreshold` (3) failures, `pool.Reset()` is called — closes all idle connections and marks checked-out ones for close-on-return, forcing fresh connections on the next `Acquire()`.
- **`HealthCheckPeriod` tightened**: `configureSharedPool` now sets `HealthCheckPeriod` to 15 s (was 30 s) so the pool self-heals within one poll interval after a Postgres restart.

## Tests added

- `TestGlobalReembedder_ConsecutiveErrorCounterResets` (unit, no DB)
- `TestGlobalReembedder_HasPoolResetOnThreshold` (source guard, no DB)
- `TestGlobalReembedder_RecoveryIntegration` (integration, skips without `TEST_DATABASE_URL`)
- `TestSharedPoolConfig_HasLifetimeSettings` (unit, no DB)
- `TestSharedPoolConfig_HealthCheckPeriod` (unit, no DB — regression guard for 15 s cap)

## Branch note

This is a re-open of the #645 fix on a clean branch. The previous branch (`fix/645-reembed-worker-hang`, PR #792) was closed because it contained a duplicate of the `anthropic.go` transport fix from PR #797 (#772), not the reembed fix. This branch contains only the correct `00337c8` cherry-pick.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/internal/db/postgres.go
+++ b/internal/db/postgres.go
@@ -60,7 +60,9 @@ func configureSharedPool(cfg *pgxpool.Config) {
 	cfg.MaxConns = 50
 	cfg.MaxConnLifetime = 30 * time.Minute
 	cfg.MaxConnIdleTime = 3 * time.Minute
-	cfg.HealthCheckPeriod = 30 * time.Second
+	// 15s health-check so the pool detects dead connections within one
+	// GlobalReembedder poll interval after a Postgres restart (#645).
+	cfg.HealthCheckPeriod = 15 * time.Second
 }
 
 // registerTypesAfterConnect registers custom type codecs that every connection
--- a/internal/db/postgres_pool_config_test.go
+++ b/internal/db/postgres_pool_config_test.go
@@ -57,3 +57,49 @@ func TestPoolConfig_LifetimeValues(t *testing.T) {
 		t.Errorf("HealthCheckPeriod: got %v, want %v", got, want)
 	}
 }
+
+// TestSharedPoolConfig_HasLifetimeSettings verifies that configureSharedPool
+// sets all connection lifetime fields to non-zero values. The shared pool is
+// used by the server process (including the GlobalReembedder); a zero
+// HealthCheckPeriod would disable proactive detection of dead connections after
+// a Postgres restart (#645).
+func TestSharedPoolConfig_HasLifetimeSettings(t *testing.T) {
+	cfg, err := pgxpool.ParseConfig("postgres://user:password@localhost:5432/engram")
+	if err != nil {
+		t.Fatalf("ParseConfig: %v", err)
+	}
+
+	configureSharedPool(cfg)
+
+	if cfg.MaxConns <= 0 {
+		t.Errorf("MaxConns: got %d, want > 0", cfg.MaxConns)
+	}
+	if cfg.MaxConnLifetime == 0 {
+		t.Error("MaxConnLifetime is zero — stale connections will never be evicted")
+	}
+	if cfg.MaxConnIdleTime == 0 {
+		t.Error("MaxConnIdleTime is zero — idle connections will never be reaped")
+	}
+	if cfg.HealthCheckPeriod == 0 {
+		t.Error("HealthCheckPeriod is zero — dead connections will not be detected proactively")
+	}
+}
+
+// TestSharedPoolConfig_HealthCheckPeriod verifies that HealthCheckPeriod is at
+// most 15 seconds for the shared pool. A shorter period means stale connections
+// (from a Postgres restart) are detected within one GlobalReembedder poll
+// interval, preventing the silent-hang described in #645.
+func TestSharedPoolConfig_HealthCheckPeriod(t *testing.T) {
+	cfg, err := pgxpool.ParseConfig("postgres://user:password@localhost:5432/engram")
+	if err != nil {
+		t.Fatalf("ParseConfig: %v", err)
+	}
+
+	configureSharedPool(cfg)
+
+	const maxAllowed = 15 * time.Second
+	if cfg.HealthCheckPeriod > maxAllowed {
+		t.Errorf("HealthCheckPeriod: got %v, want ≤ %v — too long to detect post-restart stale connections promptly (#645)",
+			cfg.HealthCheckPeriod, maxAllowed)
+	}
+}
--- a/internal/db/postgres_shared_pool_test.go
+++ b/internal/db/postgres_shared_pool_test.go
@@ -12,6 +12,9 @@ import (
 // TestSharedPoolConfig verifies configureSharedPool sets values tuned for
 // a single pool shared across all project backends: higher MaxConns to handle
 // concurrent projects, lower MinConns to avoid wasting slots on idle instances.
+// HealthCheckPeriod was tightened from 30s to 15s in #645 so the pool detects
+// dead connections within one GlobalReembedder poll interval after a Postgres
+// restart, eliminating the silent-hang regression.
 func TestSharedPoolConfig(t *testing.T) {
 	cfg, err := pgxpool.ParseConfig("postgres://user:password@localhost:5432/engram")
 	if err != nil {
@@ -29,7 +32,9 @@ func TestSharedPoolConfig(t *testing.T) {
 	if got, want := cfg.MaxConnIdleTime, 3*time.Minute; got != want {
 		t.Errorf("MaxConnIdleTime: got %v, want %v", got, want)
 	}
-	if got, want := cfg.HealthCheckPeriod, 30*time.Second; got != want {
+	// 15s: tightened from 30s in #645 so stale connections are detected within
+	// one GlobalReembedder poll interval after a Postgres restart.
+	if got, want := cfg.HealthCheckPeriod, 15*time.Second; got != want {
 		t.Errorf("HealthCheckPeriod: got %v, want %v", got, want)
 	}
 	if cfg.MaxConnLifetime == 0 {
--- a/internal/reembed/global_worker.go
+++ b/internal/reembed/global_worker.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/jackc/pgx/v5/pgxpool"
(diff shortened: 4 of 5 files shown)


petersimmons1972:
### Consolidated 4-Round Review — APPROVE WITH NOTED CONCERNS (no blockers, carry-forward findings from #792 largely resolved)

**PR:** fix(reembed): recover from stale connection after Postgres restart (#645)
**Scope:** `internal/db/postgres.go` (HealthCheckPeriod), `internal/reembed/global_worker.go` (atomic counter + pool.Reset()), tests.

---

#### Round 1 — Correctness

**Atomic counter is correct.** `atomic.Int64` with `Add(1)` on error and `Store(0)` on success is safe for concurrent access. The worker runs a single goroutine (`run()`), but `ConsecutiveErrors()` is exported for tests, and pool operations can race — the atomic is the right choice regardless.

**`pool.Reset()` call placement.** The reset fires when `consecutive >= consecutiveErrorThreshold` (3). It fires on *every* batch error above threshold, not just the first time threshold is crossed. This means if the pool stays broken, `pool.Reset()` is called on every subsequent error. That is harmless — `pool.Reset()` is idempotent — but worth acknowledging. A common alternative is to reset only when `consecutive == threshold` (not `>=`). The current approach is more aggressive but not incorrect.

**Shared-pool blast radius (carry-forward from #792 reviewers).** `pool.Reset()` closes all idle connections and marks checked-out ones for close-on-return for the *entire shared pool* — not just connections used by `GlobalReembedder`. If other callers (project backends, MCP handlers) have active transactions at the moment of reset, their connections will be closed on return, and they will establish new ones on their next `Acquire()`. This is a degraded-service event (not a crash), and it only happens after 3 consecutive errors — meaning the pool is already broken for everyone. The blast radius concern from #792 is real but the alternative (a dedicated pool for the reembedder) is a larger refactor. **The current implementation is acceptable given the failure context (pool is already broken) — this is not a blocker.** File as nice-to-have: document the shared-pool blast radius in the code comment, or link to a follow-up issue for a dedicated reembedder pool.

**HealthCheckPeriod 15s coupling (carry-forward from #792 reviewers).** The PR body explains the coupling: 15s was chosen to match the GlobalReembedder poll interval, so stale connections are detected within one poll cycle. The comment in `configureSharedPool` and the PR body both document this. The coupling is intentional and documented — not unexplained. This was a valid #792 finding; it is resolved in this PR.

**Verdict:** No correctness blockers.

---

#### Round 2 — Coverage

Tests added:
- `TestGlobalReembedder_ConsecutiveErrorCounterResets` — unit, no DB. Validates Add/Store semantics of the counter. This is a pure struct-field test; it does not call `run()`. Adequate for the counter mechanics.
- `TestGlobalReembedder_HasPoolResetOnThreshold` — source-presence guard. Reads `global_worker.go` and asserts `pool.Reset()` and `consecutiveErrorThreshold` strings are present. This is the antipattern flagged by a #792 reviewer: it tests source text, not behavior. A future refactor (e.g., renaming the method or extracting it) would break this test even if behavior is preserved. **This is a nice-to-have concern, not a blocker** — the test prevents accidental deletion but is fragile. The reviewer finding from #792 holds.
- `TestGlobalReembedder_RecoveryIntegration` — integration, skips without `TEST_DATABASE_URL`. Tests the full recovery path: seed a NULL-embedded chunk, reset the pool, start the worker, assert chunk is embedded within 25s. This is the correct integration test for the primary recovery claim. It also validates `ConsecutiveErrors()` implicitly (the worker must recover and process).
- `TestSharedPoolConfig_HasLifetimeSettings` / `TestSharedPoolConfig_HealthCheckPeriod` — unit, no DB. These guard the HealthCheckPeriod ≤ 15s invariant and the non-zero lifetime settings.

Coverage assessment: the source-presence test (`HasPoolResetOnThreshold`) covers the threshold guard at the text level. The unit counter test covers the atomic mechanics. The integration test covers the end-to-end recovery. The gap is a unit test that exercises `run()` with a mock pool that returns errors — verifying that after 3 errors, `pool.Reset()` is called on a mock. Without this, the connection between "counter reaches threshold" and "pool.Reset() is called" is only verified by the integration test (which requires a real DB). This is a nice-to-have gap.

**Verdict:** Coverage adequate for the stated fix. Source-presence test is a known antipattern — track as nice-to-have.

---

#### Round 3 — Structural

`consecutiveErrorThreshold = 3` is an unexported package-level constant. The value 3 is reasonable (not too aggressive, not too slow to recover). It is not configurable at runtime. For a background worker, this is appropriate — no need for runtime tuning of internal recovery behavior.

`ConsecutiveErrors()` exported for tests is a reasonable design. The alternative (unexported + test-only accessor in the same package) would require the test to be in the `reembed` package, which it is (`package reembed` in the test file). So `ConsecutiveErrors()` is exported unnecessarily. This is a minor style issue — nice-to-have, not a blocker.

The `run()` loop changes are readable. The `if consecutive >= consecutiveErrorThreshold` block has a clear comment linking to #645. The success reset (`g.consecutiveErrors.Store(0)`) is placed after the `if n == 0` backoff block — meaning a successful empty-queue iteration also resets the counter. This is correct: an empty queue means the DB is reachable.

`TestGlobalReembedder_RecoveryIntegration` uses `testutil.DSN(t)` to skip without `TEST_DATABASE_URL`, consistent with the rest of the integration test suite. The 25s deadline is generous relative to the 15s HealthCheckPeriod — good.

**Verdict:** No structural blockers. One nice-to-have: consider unexporting `ConsecutiveErrors()` since the test is already in the `reembed` package.

---

#### Round 4 — Zero-Context Fresh Eyes

Reading only the diff: an atomic counter is added to `GlobalReembedder`. Each batch error increments it; success resets it. After 3 consecutive errors, `pool.Reset()` is called with a warning log. `HealthCheckPeriod` is tightened from 30s to 15s with a comment explaining why. Three new tests exercise the counter, the presence of the recovery code, and an integration end-to-end.

Questions a zero-context reviewer would ask:
1. *Why 3?* Documented in the constant comment — tolerate transient single-failure noise, but not sustained failures. Reasonable.
2. *What does pool.Reset() do to concurrent callers?* Not documented in the call site comment. The comment says "forces fresh connections on the next Acquire()" but does not mention the blast radius for other callers sharing the pool. Add one sentence: "Other callers sharing this pool will also have their idle connections dropped." Nice-to-have.
3. *Why 15s and not 10s or 30s?* Documented — matches GlobalReembedder poll interval. ✅

**Verdict:** No blockers from fresh-eyes pass. One nice-to-have documentation gap at the `pool.Reset()` call site.

---

#### Re-evaluation of #792 carry-forward findings

| #792 Finding | Status in #798 |
|--------------|----------------|
| Shared-pool blast radius undocumented | Partially resolved — PR body explains it; call-site comment does not. Nice-to-have. |
| HealthCheckPeriod 15s coupling unexplained | Resolved — documented in code comment and PR body. ✅ |
| Source-presence test is an antipattern | Still present in this PR. Acknowledged nice-to-have. |
| Integration test skips without DB | Still true and appropriate — correct behavior. ✅ |

---

#### Summary

| Round | Result |
|-------|--------|
| Correctness | ✅ No blockers — blast radius concern noted but acceptable in context |
| Coverage | ✅ Adequate — source-presence test antipattern is nice-to-have |
| Structural | ✅ Clean — minor unexported-export nice-to-have |
| Zero-context | ✅ No blockers |

**APPROVE.** No `severity/blocker` findings. Nice-to-have items: (1) add one sentence to the `pool.Reset()` call-site comment about shared-pool blast radius; (2) replace `HasPoolResetOnThreshold` source-text guard with a behavior test using a mock pool; (3) consider unexporting `ConsecutiveErrors()`.

Issue #733, opened 2026-05-18
instinctllm/olla: pickModel re-runs on every Complete call — no caching


petersimmons1972:
Per prior R1 nice-to-have. `internal/instinctllm/olla.go` calls `pickModel` inside every `Complete` invocation (HTTP round-trip to list models, scoring, selection). The Python equivalent (`HaikuClient.__init__`) resolves the model once at construction.

For the instinct consolidator (one Complete per batch) the impact is small. For audit (one Complete per pattern, 50+ calls) this adds observable latency with no benefit — the available model list is stable across a single audit run.

Recommend: cache the resolved model ID in a struct field using `sync.Once` or a pointer-with-mutex, re-resolving only on miss (model disappears from Olla). Guards: reset cache on ErrBackendUnavailable from pickModel so a flapping host recovers gracefully.

petersimmons1972:
Implemented in PR #801 (commit 5df06f4 on the merged branch, now main `e195db2`). Added `sync.Once` caching for `resolvedModel` in `internal/instinctllm/olla.go` with `TestOllaPickModelCachedAfterFirstCall` as the regression guard. Closing.

Issue #732, opened 2026-05-18
cmd/instinct: LLM timeout hardcoded 30s while cmd/audit exposes --timeout flag


petersimmons1972:
Per re-review R3-N3. `cmd/instinct/main.go` hardcodes `Timeout: 30 * time.Second` in LLM Config. `cmd/audit` exposes a `--timeout` flag, making the same dependency configurable in one binary but not the other.

Asymmetric configurability of the same dependency is a maintenance hazard — a user needing a longer timeout for a slow Olla host must fork or patch.

Recommend: thread timeout through env var (`INSTINCT_LLM_TIMEOUT`) with a 30s default, mirroring the pattern used for other instinct env vars.

Issue #731, opened 2026-05-18
instinctllm/olla: skipFamilies map and HasPrefix("qwen3") check are duplicative


petersimmons1972:
`internal/instinctllm/olla.go:25-28` declares a `skipFamilies` map. Lines 141-143 also do `strings.HasPrefix(strings.ToLower(model.ID), "qwen3")`. Both filter qwen3.

Either the map is authoritative or the prefix check is; both creates future-bug surface when a third family is added. If a developer adds a new family to the map and forgets the parallel prefix branch (or vice versa), the filter silently becomes inconsistent.

Recommend: pick the map, drop the prefix check (or vice versa) and let the map be the single source of truth for skip-listed families.

Issue #726, opened 2026-05-18
E1 review nice-to-haves — cluster of style and inline-doc improvements


petersimmons1972:
**Context**

Track E1 of the instinct-python → engram-go migration added the `pattern_confidence DOUBLE PRECISION` column (commits a2a4e52, a3bd157, ff2177e on `feat/instinct-canonical`). The three-round adversarial review (correctness via superpowers:code-reviewer, coverage via rickover-validator, structural via zero-context-reviewer) surfaced 3 blockers (addressed in E1-FIX) and 13 nice-to-haves. This issue tracks the 13 nice-to-haves as a single piece of follow-up work, grouped by theme.

Reference: migration plan at `~/.claude/plans/okay-we-need-to-linear-valley.md`, Phase 1 Track E.

**Theme 1 — Wrong-type arg handling (3 findings)**

- (R1.N1) `args["pattern_confidence"].(float64)` with comma-ok silently drops non-float64 types (string, bool). Consistent with codebase pattern (`getInt` at `internal/mcp/tools.go:298-308` uses type-switch for both float64 and int). Consider extending `pattern_confidence` to accept both.
- (R2.N1) Same defect class — silent data loss when caller sends `"pattern_confidence": "high"`. No test currently covers wrong-type input asserting `IsError == true`. Proposed gate: any new optional MCP arg requires a wrong-type test case.
- (R1.N3) Missing `TestMemoryStorePatternConfidenceExplicitZero` for the 0.0 boundary case (distinct from absent). Important to lock in semantics against future refactors that might conflate `v == 0.0` with "arg absent."

**Theme 2 — Inline comments at call sites (2 findings)**

- (R3.N3) `ValidatePatternConfidence` returns error vs `ValidateImportance` silently clamping — the divergence is well-justified in `ValidatePatternConfidence`'s godoc but not visible at call sites. Add one-line comment at each call site in `tools_store.go:69` and `:359`: `// Error (not clamp) is intentional — see ValidatePatternConfidence godoc.`
- (R3.N2) `rowToMemory` and `rowToFTSResult` both scan `pattern_confidence` correctly. Optional belt-and-suspenders: add `// Column order must stay in sync with rowToMemory.` cross-reference comment to `rowToFTSResult`.

**Theme 3 — Options-struct pattern (1 finding)**

[...]

PR #801, opened 2026-05-20
refactor(instinctllm): style, dedup, timeout, caching (#726, #731, #732, #733)

petersimmons1972:
## Summary

- **#726** — Add `// Error (not clamp) is intentional` call-site comments at both `ValidatePatternConfidence` call sites in `internal/mcp/tools_store.go`. Add `TestMemoryStorePatternConfidenceExplicitZero` to lock in semantics that `pattern_confidence=0.0` is sent explicitly (distinct from absent).
- **#731** — Remove the parallel `strings.HasPrefix(model.ID, "qwen3")` check from `pickModel`; `skipFamilies` map (keyed on `olla.Family`) is now the single source of truth for the deny-list. Add `TestOllaSkipFamiliesMapIsAuthoritative` (subtests: qwen3, qwen3moe).
- **#732** — Replace hardcoded `30*time.Second` LLM timeout in `cmd/instinct` with `envDuration("INSTINCT_LLM_TIMEOUT", 30*time.Second)`. Add `envDuration()` helper and `TestEnvDuration`. Note: `internal/config` (#794) not yet merged; inline default pattern used.
- **#733** — Cache `pickModel` result per client using `sync.Once` (stored as `modelID` on `ollaClient`). Add `resolvedModel()` and `resetModel()` methods. Cache is cleared on transport/HTTP errors so flapping hosts recover. Add `TestOllaPickModelCachedAfterFirstCall` verifying discovery endpoint is called exactly once for N Complete calls.

Note: #731 and #733 both touched `olla.go`/`olla_test.go` and landed in the same commit (5df06f4).

## Tests added

- `TestMemoryStorePatternConfidenceExplicitZero` (cmd/instinct)
- `TestOllaSkipFamiliesMapIsAuthoritative` (internal/instinctllm, 2 subtests)
- `TestOllaPickModelCachedAfterFirstCall` (internal/instinctllm)
- `TestEnvDuration` (cmd/instinct, 4 subtests)

## Verification

`go build ./...`, `go vet ./...`, and `go test ./internal/instinctllm/... ./internal/mcp/... ./cmd/instinct/... -count=1 -race -short` all pass clean. Pre-existing build failures in other packages (unrelated, present on main) are not caused by this branch.

### DIFF
--- a/cmd/instinct/main.go
+++ b/cmd/instinct/main.go
@@ -160,6 +160,18 @@ func envInt(key string, def int) int {
 	return def
 }
 
+// envDuration returns the time.Duration value of the named environment variable
+// (parsed via time.ParseDuration), or def when unset, empty, or not parseable.
+func envDuration(key string, def time.Duration) time.Duration {
+	if v := os.Getenv(key); v != "" {
+		if d, err := time.ParseDuration(v); err == nil {
+			return d
+		}
+		slog.Warn("instinct: ignoring unparseable duration env var", "key", key, "value", v)
+	}
+	return def
+}
+
 // loadAndRotate reads the buffer JSONL. Returns empty if file missing or
 // line count < minEvents. On success renames to .processed and returns events + new path.
 type bufferLoadOutcome int
@@ -735,7 +747,10 @@ func run(ctx context.Context, cfg config) error {
 		Backend:  envOr("LLM_BACKEND", "anthropic"),
 		APIKey:   cfg.anthropicKey,
 		Endpoint: cfg.haikuEndpoint,
-		Timeout:  30 * time.Second,
+		// Per-inference timeout: configurable via INSTINCT_LLM_TIMEOUT env var
+		// (e.g. "60s", "2m") so operators with slow Olla hosts can increase it
+		// without rebuilding.  Mirrors the --timeout flag in cmd/audit.
+		Timeout: envDuration("INSTINCT_LLM_TIMEOUT", 30*time.Second),
 	})
 	var patterns []Pattern
 	if err != nil {
--- a/cmd/instinct/main_test.go
+++ b/cmd/instinct/main_test.go
@@ -65,6 +65,39 @@ func TestLoadConfig_Defaults(t *testing.T) {
 	}
 }
 
+// TestEnvDuration verifies the envDuration helper used for INSTINCT_LLM_TIMEOUT.
+// Ref: #732.
+func TestEnvDuration(t *testing.T) {
+	t.Run("returns_default_when_unset", func(t *testing.T) {
+		t.Setenv("ENVTEST_DUR_MISSING", "")
+		got := envDuration("ENVTEST_DUR_MISSING", 30*time.Second)
+		if got != 30*time.Second {
+			t.Errorf("envDuration() = %v, want 30s (default)", got)
+		}
+	})
+	t.Run("parses_valid_duration", func(t *testing.T) {
+		t.Setenv("ENVTEST_DUR_VALID", "2m")
+		got := envDuration("ENVTEST_DUR_VALID", 30*time.Second)
+		if got != 2*time.Minute {
+			t.Errorf("envDuration() = %v, want 2m", got)
+		}
+	})
+	t.Run("returns_default_on_invalid_value", func(t *testing.T) {
+		t.Setenv("ENVTEST_DUR_BAD", "notaduration")
+		got := envDuration("ENVTEST_DUR_BAD", 30*time.Second)
+		if got != 30*time.Second {
+			t.Errorf("envDuration() = %v, want 30s (default on bad input)", got)
+		}
+	})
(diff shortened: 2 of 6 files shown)


petersimmons1972:
### Consolidated 4-Round Review — APPROVE (0 blockers, 2 nice-to-haves)

Reviewed branch `refactor/4b-instinctllm-cluster` against `main`. All new tests pass (`go test ./internal/instinctllm/... ./cmd/instinct/... -count=1 -race`). No blockers.

---

#### Round 1 — Correctness

**#731 — `skipFamilies` dedup:** `strings.HasPrefix("qwen3")` is fully removed; `strings` import retained for the `stripThinkingTags` usage at line 261 (correct). The map is now the single source of truth. No regression: `TestOllaSkipFamiliesMapIsAuthoritative` directly proves both qwen3 and qwen3moe are skipped via family field alone. ✅

**#733 — `sync.Once` caching:** Primitive choice is correct for single-goroutine callers (`cmd/instinct` calls `Detect` sequentially; `cmd/audit` calls `Complete` sequentially — confirmed by grep). The `resetModel` struct-assignment (`c.modelOnce = sync.Once{}`) would be a data race if `ollaClient` were shared across goroutines, but it isn't in the current callers. The comment in the struct says "model list is stable within a single run" — this is accurate. However, `resetModel` has **0% test coverage** (confirmed via `go tool cover`): the transport-error and non-200 paths that call `resetModel()` are not exercised by any test. The re-discovery-after-failure semantic is documented in the godoc but unverified. Flagged below as nice-to-have.

**#732 — `envDuration` edge cases:** `time.ParseDuration` rejects negative values only if they're truly invalid format — it actually accepts `-1s` and returns a negative `time.Duration`. `envDuration` would silently pass a negative timeout to the client, and `ollaClient` uses it directly in `context.WithTimeout` — which treats a non-positive timeout as immediate cancellation. A zero or negative `INSTINCT_LLM_TIMEOUT` would cause every call to fail with a context deadline exceeded. `TestEnvDuration` has no subtest for negative or zero values. Flagged below as nice-to-have (operator misconfiguration, not a correctness bug in normal use).

**#726 — Call-site comments + explicit-zero test:** Comment placement at both `ValidatePatternConfidence` call sites in `tools_store.go` is correct and symmetric. `TestMemoryStorePatternConfidenceExplicitZero` correctly distinguishes `pattern_confidence=0.0` from absent field. ✅

---

#### Round 2 — Coverage

| Package | Coverage | Gate (55%) |
|---|---|---|
| `internal/instinctllm` | 83.3% ✅ | Above 55% |
| `cmd/instinct` | 78.3% ✅ | Above 55% |

New functions by coverage:
- `resolvedModel` — 100% ✅
- `resetModel` — **0%** ⚠️ (nice-to-have, not blocker — not an exported API)
- `envDuration` — 100% ✅

Missing: no test covers the "Complete fails → resetModel → next call re-discovers" recovery path. `TestOllaPickModelCachedAfterFirstCall` only tests the happy path (discovery called once). The advertised flapping-host recovery has no test.

---

#### Round 3 — Structural

**`envDuration` placement:** Correctly placed in `cmd/instinct/main.go` alongside `envOr`/`envInt`. PR body documents that `internal/config` (#794) was not merged at branch time — inline pattern was the right call, and it matches existing convention. ✅

**`resolvedModel`/`resetModel` naming:** Clear and conventional. Lowercase = unexported, correct for a private implementation method. ✅

**Struct comment for `modelOnce`:** The comment says "running it N times on the same Olla host always returns the same model (model list is stable within a single run)" — this is an assumption about Olla's behavior, not a Go-level guarantee, but it's appropriate to document. ✅

**`sync` import:** Correctly added; was not previously imported. ✅

---

#### Round 4 — Zero-Context

PR title matches diff: style, dedup (#731), timeout (#732), caching (#733), and the `#726` call-site comments land in exactly 6 files. Commit log matches: 3 commits (d9ccf79, 5df06f4, a54e442), each closing one issue. `#730` (rename instinctllm) is absent from the diff — correctly excluded as a separate concern. PR body accurately describes the `#794` dependency. ✅ Branch is `MERGEABLE` (no conflicts with main). ✅

---

#### Nice-to-Have Findings (non-blocking)

**N1 — `resetModel` 0% coverage** (`severity/nice-to-have`): Add a test that verifies the cache is cleared after a transport or non-200 error so the next call re-discovers. Suggested approach: use `httptest.Server` that fails once then succeeds, verify discovery is called twice.

**N2 — `envDuration` negative/zero guard** (`severity/nice-to-have`): `time.ParseDuration("-1s")` succeeds and returns a negative duration; `context.WithTimeout` with a negative deadline cancels immediately. Consider adding a `d <= 0` guard in `envDuration` that logs a warning and returns the default, or add a subtest documenting the current behavior as intentional.

---

**Verdict: APPROVE.** 0 blockers. 2 nice-to-haves filed as tracking items. Safe to merge.

PR #802, opened 2026-05-20
fix(mcp): reject out-of-range importance in handleMemoryCorrect

### LINKED ISSUE #789: handleMemoryCorrect silently clamps importance; should validate symmetrically with store path
PR #786 narrowed the REST validator for store paths to 0-4, returning errors for out-of-range importance. handleMemoryCorrect (the update path) still silently clamps out-of-range values without surfacing an error. Asymmetric — store rejects, update silently corrupts. Apply the same validation in handleMemoryCorrect, or document the intentional asymmetry.

petersimmons1972:
## Summary

- `handleMemoryCorrect` was silently clamping `importance` values outside [0, 4] via `types.ValidateImportance`, while `handleMemoryStore` and `handleMemoryStoreBatch` validate and return an error. This asymmetry allows callers to silently corrupt stored importance through the update path (issue #789).
- Replaced the silent clamp with the same range check used on the store path, returning a descriptive MCP `ToolResultError` for values outside `[minImportanceValue, maxImportanceValue]`.
- Added `TestHandleMemoryCorrect_RejectsOutOfRangeImportance` as the regression guard — tests -1, 5, 10, 100 all confirmed rejected.

## Implementation notes

Single implementation approach (mirror store pattern) — no advisory gate needed per coordinator brief. Constants `minImportanceValue` / `maxImportanceValue` already defined in `tools_store.go` and reused.

The fix removes the `types.ValidateImportance` call entirely from the correct path. That function (a silent clamp) remains in `types` for callers that intentionally want clamping; it is no longer appropriate here.

## Test plan

- [x] `TestHandleMemoryCorrect_RejectsOutOfRangeImportance` — 4 subtests (importance=-1, 5, 10, 100), all PASS
- [x] `make test` — all packages pass
- [x] `go vet ./...` — clean

Closes #789

### DIFF
--- a/cmd/audit/report.go
+++ b/cmd/audit/report.go
@@ -76,7 +76,7 @@ func envOr(key, fallback string) string {
 // errNoEngram is returned when no Engram credentials can be resolved from
 // flags, environment variables, or the developer config file.
 var errNoEngram = errors.New(
-	"Engram credentials not found: set ENGRAM_URL and ENGRAM_TOKEN env vars, " +
+	"engram credentials not found: set ENGRAM_URL and ENGRAM_TOKEN env vars, " +
 		"or pass -engram and -token flags",
 )
 // resolveEngram returns the Engram base URL and bearer token to use.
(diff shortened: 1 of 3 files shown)


petersimmons1972:
S2 error-shape normalization (handleMemoryStore/handleMemoryStoreDocument validation errors now return mcpgo.NewToolResultError) applied via #809 commit 223fe77.

Issue #751, opened 2026-05-20
lme: project auto-cleanup deletes ingested projects before cross-experiment reuse is possible


petersimmons1972:
## Problem

Exp 13 planned to reuse ingested projects from Exp 10 but they had been auto-deleted by the \`runRun\` cleanup path. The only opt-out is \`--no-cleanup\` (a double-negative flag name), which must be passed at run time. If the operator forgets the flag, the projects are gone with no recovery path.

## Root Cause

\`cmd/longmemeval/run.go:170-175\` deletes each item's Engram project immediately after the run phase completes, unless \`cfg.NoCleanup\` is true. The default is \`NoCleanup: false\` (i.e., delete). There is no preservation flag that communicates positive intent.

## Proposed Fix

**Advisory gate recommendation: Option B** — add a \`--preserve-projects\` flag as an alias/replacement for \`--no-cleanup\`. Keep delete as the default to avoid breaking existing camp pipeline scripts. Document the flag prominently in the help output with a note: "Use --preserve-projects if you plan to reuse ingested data in a follow-up experiment."

The existing \`--no-cleanup\` flag can remain for backwards compatibility but should be documented as deprecated in favour of \`--preserve-projects\`.

## References

- \`cmd/longmemeval/run.go:170-175\` — cleanup path
- \`cmd/longmemeval/main.go:111\` — flag registration
- \`results/lme-camp-state.json\` — Exp 10/13 reuse context
- Design doc: \`docs/proposed-fixes/<NNN>-project-cleanup-preserve-flag.md\`

PR #807, opened 2026-05-20
fix(lme): cleanup-policy enum with auto default — preserve cross-experiment projects (#751)

petersimmons1972:
## Summary

- Introduces `--cleanup-policy {auto,always,never}` (default: `auto`) replacing the boolean `--no-cleanup` flag
- `auto` deletes only projects whose name matches the `lme-{runID}-*` prefix created by the current invocation; externally ingested / cross-experiment projects are preserved
- `always` restores pre-fix unconditional-delete behavior; `never` preserves all
- `--no-cleanup` is deprecated but honored as `never`-alias; emits `WARN: --no-cleanup is deprecated` to stderr
- Provenance detection: pure in-memory via `projectName()` prefix (`lme-{runID}-{qid}`) — no checkpoint schema changes required

## Advisory reference

Implemented per opus-advisor Path S recommendation (cleanup-policy enum, auto default, in-memory prefix provenance).

## Test plan

- [ ] `TestCleanupPolicy_AutoPreservesExternalProject` — auto preserves project with mismatched runID prefix
- [ ] `TestCleanupPolicy_AutoDeletesIngestedProject` — auto deletes project matching this run's prefix
- [ ] `TestCleanupPolicy_AlwaysDeletes` — always policy deletes every project
- [ ] `TestCleanupPolicy_NeverPreserves` — never policy preserves every project
- [ ] `TestNoCleanupFlag_DeprecatedAliasForNever` — --no-cleanup registered, deprecation warning present
- [ ] `TestCleanupPolicyFlag_DefaultIsAuto` — --cleanup-policy default is CleanupPolicyAuto
- [ ] `TestCleanupPolicyStructuralGuard` — run.go uses shouldCleanupProject(), not cfg.NoCleanup
- [ ] `go test -race ./cmd/longmemeval/...` passes
- [ ] `go vet ./...` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to engram-go are documented here.
 
 ## [Unreleased] — v3.3.0
 
+### Bug Fix — lme cleanup-policy (#751)
+
+- **`--cleanup-policy` enum (default: `auto`):** The `lme run` stage now ships a three-value cleanup policy instead of the old unconditional delete. `auto` (new default) deletes only projects whose name matches the `lme-{runID}-*` prefix created by the current invocation; externally ingested or cache-reused projects are preserved. `always` restores the pre-fix unconditional-delete behavior. `never` preserves everything (useful for cross-experiment reuse such as Exp 10 → Exp 13).
+- **`--no-cleanup` deprecated:** The old boolean flag remains registered for backward compatibility and is silently coerced to `--cleanup-policy=never`. A `WARN: --no-cleanup is deprecated` message is emitted to stderr on use.
+- **Migration note:** Existing camp pipeline scripts that relied on `--no-cleanup` continue to work unchanged. Scripts that want the old unconditional-delete behavior must pass `--cleanup-policy=always` explicitly going forward.
+
+---
+
+## [Unreleased] — v3.3.0
+
 ### Reliability (PR #611-fix2)
 
 - **Async embed on store (architecturally enforced):** `memory_store` and `memory_store_batch` return after the DB write completes (~10ms), regardless of embed pool state. Chunks are stored with NULL embeddings and the existing reembed worker backfills them asynchronously. New Prometheus counter `engram_store_embed_async_total` tracks call volume on the async path. Rollback: set `ENGRAM_STORE_EMBED_MODE=sync` to restore inline embedding without redeploying.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,17 @@ All notable changes to engram-go are documented here.
 
 ## [Unreleased] — v3.3.0
 
-### Bug Fix — lme cleanup-policy (#751)
+### Bug Fix — lme cleanup-policy (#751, PR #807)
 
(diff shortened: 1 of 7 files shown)


petersimmons1972:
Round 2 QA fixes applied: B2, B3, S3, S7, S9.

**B2** (BLOCKER):  now validated against  immediately after ; unknown values return `invalid --cleanup-policy %q: must be one of auto|always|never` before any downstream check. Tests: `TestCleanupPolicy_UnknownValueRejected`, `TestCleanupPolicy_ValidValuesAccepted`.

**B3** (BLOCKER): `--no-cleanup` + explicit non-default `--cleanup-policy` now returns `conflicting flags: ...` instead of silently overwriting. Coercion-to-never still happens when CleanupPolicy is the default (auto), i.e. user only passed `--no-cleanup`. Tests: `TestNoCleanup_ConflictsWithExplicitPolicy`, `TestNoCleanup_AloneCoercesToNever`.

**S3** (SERIOUS): slog not yet wired in `cmd/longmemeval` (lives in `internal/` only) — scoped down per brief. Added `TODO(#807-S3)` comment referencing structured-logging follow-up. `log.Printf` already writes to stderr via stdlib log.

**S7** (SERIOUS): `newRunID()` expanded from 3 bytes (24-bit / ~16M) to 8 bytes (64-bit / ~18E) to eliminate prefix-match collision risk on shared infra.

**S9** (SERIOUS): Per-item preserve logs eliminated. Workers now send preserved-project names to a `preservedCh`; `runRun` drains it after `wg.Wait()` and emits one summary: `INFO run: preserved N project(s) under cleanup-policy=X (sample: [...])`.

Commit: 235f8e9 · `go test -race ./cmd/longmemeval/...` ✅ · `go vet ./...` ✅ · `golangci-lint` 0 issues ✅

petersimmons1972:
Round 3 fixes pushed (9bf9e2c):

**R2-B2 deadlock** — replaced `preservedCh` buffered channel with mutex-protected `[]string` slice (`preservedLog` struct). Channel blocked senders when preserved count > `cfg.Workers*2`; slice has no capacity limit. Tests: `TestPreservedLog_NoDeadlockOnHighCount` (N=100 goroutines, 5s timeout), `TestPreservedLog_NamesReturnsCopy`.

**R2-B4 RunID** — `newRunID()` confirmed at 8 bytes (16 hex chars, 64-bit entropy). Was already applied in working tree; committed cleanly.

**S9 token** — `cleanup-summary` token confirmed at line start. Structural guard `TestCleanupSummary_TokenIsCleanupSummary` verifies token and rejects old `INFO run: preserved` sentinel.

**CHANGELOG** — Deduplicated duplicate `## [Unreleased] — v3.3.0` header; v3.2.0 renamed to `## [v3.2.0]` (released); R2-B2, R2-B3, S7, S9 entries present under #807 block.

**Test coverage** — `TestCleanupPolicy_UnknownValueRejected` already included `""` empty-string case. All tests pass: `go test -race ./cmd/longmemeval/... → ok (6.0s)`. `go vet` clean. `golangci-lint` 0 issues.

PR #808, opened 2026-05-20
feat(lme): add per-backend flock contention guard

### LINKED ISSUE #749: lme: no contention guard for parallel experiments against single vLLM endpoint
## Problem

Multiple \`longmemeval run\` processes can be launched in parallel against the same oblivion vLLM endpoint with no coordination. During the LME-on-a-Budget campaign Phase 0, multiple scoring jobs auto-started simultaneously, causing GPU contention that had to be killed manually.

## Root Cause

\`cmd/longmemeval/run.go\` has no mechanism to detect or prevent concurrent runs against the same \`--llm-url\` endpoint. \`runRun()\` fires immediately with \`cfg.Workers\` goroutines; there is no lockfile, semaphore, or advisory check.

## Proposed Fix

Add a \`--exclusive-backend\` flag that creates a lockfile at \`~/.cache/longmemeval/<hash-of-llm-url>.lock\`. If the lock is held, print a clear error and exit 1. Include 4h TTL so crashed processes don't permanently block. Alternatively, document the antipattern as a WARNING in the --help output.

## References

- \`cmd/longmemeval/run.go\` — \`runRun()\` entry point
- \`results/lme-camp-report.md\` — "Multiple scoring jobs auto-started; killed manually due to resource contention"
- Design doc: \`docs/proposed-fixes/<NNN>-parallel-experiment-contention-guard.md\`

petersimmons1972:
## Summary

- Parallel `longmemeval run` invocations against the same vLLM/Infinity backend caused GPU contention (Issue #749). This PR adds a PID-liveness lockfile guard per the opus-advisor Path X recommendation.
- Lock is `flock(LOCK_EX|LOCK_NB)` on `$XDG_RUNTIME_DIR/lme/backend-<sha256[:12]>.lock` (fallback `/tmp/lme`). On `EWOULDBLOCK`, checks liveness via `kill(pid,0)`: dead PID → reclaim with WARN log; live PID → exit 2 with descriptive error.
- New flags: `--exclusive-backend` (default true), `--no-exclusive-backend`, `--backend-lock-dir <path>`.

## Implementation

Implemented per **opus-advisor Path X** recommendation: local `flock` on lockfile keyed by `sha256(normalizedURL)[:12]`. PID-liveness staleness (not wall-clock TTL). Lock file contains `<pid>\n<start_unix>\n<invocation_id>`. Acquired in `runRun` before any work begins; released via `defer`.

URL normalization: trim trailing slash, lowercase host.

## Test plan

- [x] `TestBackendLockNormalization` — URL normalizer strips trailing slashes, lowercases host
- [x] `TestLockFileName` — lockfile name is sha256[:12] of normalized URL
- [x] `TestBackendLock_Contention` — second goroutine exits with code 2 within 100ms of live holder
- [x] `TestBackendLock_StaleLock` — dead PID reclaimed with WARN log, second acquirer succeeds
- [x] `TestBackendLock_NoExclusiveBackend` — `--no-exclusive-backend` bypasses all locking
- [x] `TestBackendLock_DefaultLockDirFallback` — falls back to `/tmp/lme` when `XDG_RUNTIME_DIR` unset
- [x] `TestBackendLock_XDGRuntimeDirUsed` — uses `$XDG_RUNTIME_DIR/lme` when set
- [x] `go test -race ./cmd/longmemeval/...` — all 50 tests pass, no races
- [x] `go vet ./...` — clean
- [x] `golangci-lint run ./cmd/longmemeval/...` — 0 issues

Closes #749


petersimmons1972:
Round 2 QA fixes applied: B4, B5, S1, S3, S5, S6.

**Commit:** 352c1c4

**B4 (BLOCKER)** — pid=0 deadlock: `isProcessAlive(0)` now returns false explicitly (guards all pid<=0). WriteAt failure on lock write releases flock, removes lock file, and returns error — no zombie lock.

**B5 (BLOCKER)** — rawURL credential leak: `redactURL()` helper added (strips user-info via `net/url`; sha256[:12] fallback for unparseable URLs). All `rawURL` occurrences in `fmt.Errorf`/`log.Printf` replaced.

**S1 (SERIOUS)** — O_TRUNC TOCTOU race: stale-reclaim path now opens without O_TRUNC, acquires `LOCK_EX` flock first, then truncates+writes under the lock.

**S3 (SERIOUS)** — slog not wired in `cmd/longmemeval/`; kept `log.Printf` per scope-down rule. Added `TODO(S3):` comments on all four log sites.

**S5 (SERIOUS)** — exit code collision: contention exit changed 2 → 75 (EX_TEMPFAIL). `ExitCodeLockContention = 75` const exported. Both contention tests updated to assert exit 75. `printUsage()` documents exit 75 semantics.

**S6 (SERIOUS)** — docs: CHANGELOG.md v3.3.0 entry added; `docs/lme-benchmark-learnings.md` §9 updated with guard description, override instructions, and example error output.

**New tests:** TestBackendLock_PidZeroReclaim, TestRedactURL, TestBackendLock_ContendingExitCode75, TestBackendLock_StaleReclaimAtomicity.

**Result:** `go test -race ./cmd/longmemeval/...` — PASS. `golangci-lint` — 0 issues. `go vet` — clean.

petersimmons1972:
Round 3 fixes: R2-B1 exit code (return ExitCodeLockContention, not 2), R2-B3 query-string stripping in redactURL, sha256 fallback replaced with fixed placeholder, normalizeBackendURL comment corrected, 6x TODO(S3) anchored to #808-S3, manual lock-recovery runbook added to §9 of lme-benchmark-learnings.md. All tests pass (race detector), go vet clean, golangci-lint 0 issues. Commit: 4ba2cc7.

petersimmons1972:
**Round 4 cleanup: fragment leak fix + proposal doc annotation (85f8dbc)**

**Task 2 — redactURL fragment leak fixed:**
`redactURL` was not explicitly clearing `u.Fragment`, so URLs like `http://host/v1#token=abc` would leak the fragment into error messages. Added `u.Fragment = ""` after `u.RawQuery = ""`. Updated the doc comment — the old claim "Fragment is dropped by net/url.String() when RawQuery is cleared" was inaccurate (fragment is independent of query in net/url). New comment reads "strip userinfo, query, and fragment". Added `R2-B4` test case: `http://host/v1#token=abc` → `http://host/v1`.

**Task 3 — proposal doc annotated:**
`docs/proposed-fixes/749-parallel-experiment-contention-guard.md` now has a SUPERSEDED banner at the top noting implementation in PR #808 with exit code 75. Status line updated to `Implemented (PR #808)`.

Both changes in commit `85f8dbc` on `fix/issue-749`. Tests pass (`go test -race ./cmd/longmemeval/... ./internal/mcp/...`), `go vet` clean.

PR #809, opened 2026-05-20
fix(mcp): recalculate valid_from on memory_correct when tags change (#765)

### LINKED ISSUE #765: nice-to-have: memory_correct does not recalculate valid_from when date: tags change

petersimmons1972:
## Summary

Fixes the silent stale-date bug in `memory_correct`: when the `tags` argument is present, `valid_from` is now always recalculated from the resulting tag set (Path α — opus-advisor RECOMMENDATION). Passing `tags: []` or a tag set with no `date:` tag now clears `valid_from` to NULL, matching the behaviour of the store path.

Closes #765

## Root Cause

`UpdateMemory` used an "only-promote, never-nullify" policy: it would only set `valid_from` when `ParseDateTag` found a `date:` tag in the **new** set, silently preserving the old value when the `date:` tag was removed. The fix replaces the conditional promotion with an unconditional `types.ParseDateTag` assignment whenever `tags != nil`.

## Advisory

Binding opus-advisor RECOMMENDATION: **Path α — always recalculate from tags when the `tags` arg is present.** Callers that want to preserve an existing `valid_from` must omit `tags` entirely.

## Changes

- `internal/db/postgres_memory.go` — unconditional `valid_from` recalc when `tags != nil`
- `internal/db/postgres_valid_from_test.go` — renamed `TestUpdateMemory_PreservesValidFromWhenTagsHaveNoDate` → `TestUpdateMemory_ClearsValidFromWhenTagsHaveNoDate`; updated assertion to expect NULL
- `internal/mcp/tools_store_test.go` — added `validFromCorrectBackendV2` + introduced `CorrectPatch` struct; five Path α test assertions:
  - **(a)** store with `date:2025-01-01` tag sets `valid_from`
  - **(b)** `memory_correct` with new `date:2025-06-01` tag updates `valid_from`
  - **(c)** `memory_correct` with `tags: []` clears `valid_from` to NULL
  - **(d)** `memory_correct` omitting `tags` preserves existing `valid_from`
  - **(e)** history record is created for each correct operation

## Test results

```
ok  github.com/petersimmons1972/engram/internal/mcp  (race)
ok  github.com/petersimmons1972/engram/internal/db   (race)
go vet ./... — clean
```

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -253,9 +253,10 @@ memory_correct(
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
 | `memory_id` | string | yes | ID of the memory to update |
+| `project` | string | yes | Project namespace the memory belongs to; must match the project used at store time |
 | `new_content` | string | no | Replacement content; omit to leave content unchanged |
 | `tags` | array of strings | no | Replacement tag set; see tag-and-valid_from rules below |
-| `importance` | integer 0–4 | no | Replacement importance value; must be 0–4 |
+| `importance` | integer 0–4 | no | Replacement importance value; must be 0–4. Out-of-range returns a tool-result error (`isError: true`) with message `importance must be 0-4`. Symmetric with `memory_store` validation. |
 
 **Tag and valid_from rules (issue #765):**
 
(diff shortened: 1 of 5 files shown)


petersimmons1972:
Round 2 QA fixes applied: B1 (MCP JSON deserialization tests, advisory-gated), S2 (error-shape normalization), S6 (docs/tools.md), F5 (comment hygiene). Commit: 223fe77. All tests pass, 0 lint issues.

petersimmons1972:
Round 3 fixes: R2-B5 project param, importance error shape, SDK callout, test renames. Commit 5dc2d8a. All tests pass (race-clean). Follow-up issue #821 filed for missing handleMemoryHistory assertion in _E test.

petersimmons1972:
**Task 1 verdict: code was clamping — fix applied (891d6a9)**

The Skeptical SE review was correct. `handleMemoryCorrect` at line 379 was calling `types.ValidateImportance(int(v))` which silently clamps out-of-range values (godoc explicitly says "clamps n to the valid importance range [0, 4]"). The PR description and docs claimed it errors, but the code did not.

**Fix:** Replaced the `ValidateImportance` call with an explicit `[0, 4]` range check mirroring the store and store_batch paths, returning `mcpgo.NewToolResultError` on violation. Added `TestMemoryCorrect_ImportanceOutOfRange` with sub-cases for -1 and 5 as a regression guard.

Commit `891d6a9` on `fix/issue-765` — all tests pass.

PR #841, opened 2026-05-21
fix(lme): route --scratch-ttl through /quick-store REST body (#837)

### LINKED ISSUE #837: lme: --scratch-ttl requires direct DB access; must route through MCP instead
Follow-up to #754 (PR #835, merged 2026-05-20).

## Problem

The `--scratch-ttl` flag added in #835 sets project TTL by writing directly to the `project_ttl` table via `--database-url`. This bypasses Engram's MCP layer and assumes the LME client has Postgres access — which it does NOT in any deployed topology.

LME runs against the Engram MCP endpoint at e.g. `http://127.0.0.1:8788`. The Engram server is the only thing with PG access. The flag is functionally dead when LME and Engram don't share a host/DB — which is the normal case.

## Reproduction

Run `lme run --scratch-ttl 168h` without setting `--database-url`. The ingest stage logs a WARN "scratch-ttl: no database-url; projects will be durable" and proceeds without writing TTL rows. Projects accumulate; the feature does nothing.

## Fix shape

TTL must be set through Engram's MCP layer, not via a direct DB connection on the client. Two reasonable approaches:

1. **Thread `expires_at` through existing tool args** — extend `memory_store`, `memory_store_document`, `memory_store_batch` to accept optional `expires_at`. LME passes `now + scratch_ttl` per call.

2. **New MCP tool `memory_set_project_ttl`** — explicit one-shot per project, called once per ingest run. Cleaner separation but adds a tool to the surface.

Option 1 is simpler and matches the existing pattern (fields are properties of the memory/project, set at creation).

## Acceptance

- `lme run --scratch-ttl 168h` against a remote Engram (no shared DB) sets TTL on the created projects
- `lme prune` against the same remote works without a direct DB connection (already does — uses MCP query)
- The `--database-url` flag for the scratch-ttl path is removed or repurposed

## Context

Discovered 2026-05-20 while launching the post-PR-#805 LME baseline run. Flag was omitted to avoid the WARN noise; benchmark proceeds with durable projects (same as camp-13 baseline). Not blocking the benchmark itself.

petersimmons1972:
## Summary

- Extends `POST /quick-store` body to accept optional `expires_at` (RFC3339)
- Server calls `SetProjectTTL` after a successful store when `expires_at` is present; failure is WARN-logged, not propagated
- `RestClient.QuickStore` gains `expiresAt *time.Time` final parameter; nil omits the field from the body
- `ingestOne` computes `now + ScratchTTL` and passes it to every `QuickStore` call when `ScratchTTL > 0`
- Removes `ttlStamper` interface, direct `db.NewPostgresBackend` block, `DatabaseURL` field, and `--database-url` flag entirely

Closes #837.

## Test plan

- `TestQuickStoreHandler_ExpiresAt_*` (4 tests) — server-side TTL stamping, past-timestamp rejection, absent field, error tolerance
- `TestRestClient_QuickStore_ExpiresAt_*` (2 tests) — body serialization with and without `expiresAt`
- `TestIngestOne_ScratchTTL_*` (2 tests) — `expires_at` present when TTL > 0, absent when TTL == 0
- Full suite `go test ./... -race` — no regressions

### DIFF
--- a/cmd/longmemeval/ingest.go
+++ b/cmd/longmemeval/ingest.go
@@ -11,42 +11,12 @@ import (
 	"sync"
 	"time"
 
-	"github.com/petersimmons1972/engram/internal/db"
 	"github.com/petersimmons1972/engram/internal/longmemeval"
 )
-
-// ttlStamper writes project_ttl rows during ingest (#754). It is a tiny
-// interface so the ingest worker can be unit-tested with a fake.
-type ttlStamper interface {
-	SetProjectTTL(ctx context.Context, project string, createdAt time.Time, expiresAt *time.Time) error
-}
 
 func runIngest(cfg *Config) {
 	items := loadItems(cfg.DataFile)
 	ckptPath := filepath.Join(cfg.OutDir, "checkpoint-ingest.jsonl")
-
-	// #754: if a scratch TTL is in effect and a DSN is configured, open a
-	// PostgresBackend so each per-question project gets a project_ttl row
-	// stamped at first-store time. Failure to connect downgrades to a WARN
-	// rather than failing ingest — a missing TTL row makes the project durable
-	// (the operator backfill query is documented in the migration header).
-	var stamper ttlStamper
-	if cfg.ScratchTTL > 0 && cfg.DatabaseURL != "" {
-		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-		defer cancel()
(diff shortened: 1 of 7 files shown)


petersimmons1972:
Closing — commits 44b9c58, ac89387, b0ad096 landed directly on main. The fix is already live.

PR #953, opened 2026-06-01
fix(db): canonicalize legacy bge-m3 embedder metadata

### LINKED ISSUE #911: hygiene: stale embedder_name metadata in production — correctness depends on alias map never being removed
## Summary

21 production projects in the `project_meta` store carry `embedder_name='bge-m3-Q8_0.gguf'` (affecting 4,179 rows). These values were written before the llama.cpp `--alias` flag was standardized.

Production correctness is **safe today** — the alias map at `engine.go:1017` resolves `bge-m3-Q8_0.gguf` → `BAAI/bge-m3` at query time. However, this means production is silently depending on that alias map never being removed or refactored away.

## Risk

If the alias map is simplified, removed, or bypassed in a future refactor, those 4,179 rows will be silently mismatched and vector recall will degrade without an obvious cause.

## Recommended Fix

One-time canonicalization (idempotent, low-risk, no schema change):

```sql
UPDATE project_meta
SET value = 'BAAI/bge-m3'
WHERE key = 'embedder_name'
  AND value = 'bge-m3-Q8_0.gguf';
-- Expected: ~4,179 rows updated across 21 projects
```

> **Requires founder approval before execution** — this is a production DB write.

After execution, the alias map entry for `bge-m3-Q8_0.gguf` can be retained for safety but is no longer load-bearing.

## Evidence

- Root-cause sweep 2026-05-28, branch `fix/olla-embed-priority`
- `SELECT COUNT(*) FROM project_meta WHERE key='embedder_name' AND value='bge-m3-Q8_0.gguf'` → 4,179
- Alias map: `engine.go:1017`

## Cross-references

- engram-go#910 — reembed crash-loop (original source of the wrong names)

petersimmons1972:
## Summary
- add idempotent DB migration to canonicalize legacy `project_meta.embedder_name` values from `bge-m3-Q8_0.gguf` to `BAAI/bge-m3`
- remove hidden runtime dependency on alias-map preservation for existing rows
- keep runtime aliases as safety net, but no longer load-bearing for migrated metadata

## Validation
- `go test ./...`

Closes #911

### DIFF
--- a/internal/db/migrations/024_canonicalize_embedder_name_bge_m3.sql
+++ b/internal/db/migrations/024_canonicalize_embedder_name_bge_m3.sql
@@ -1,0 +1,10 @@
+-- 024_canonicalize_embedder_name_bge_m3.sql
+--
+-- Canonicalize historical llama.cpp GGUF model IDs in project_meta so
+-- embedder_name metadata no longer depends on runtime alias maps.
+--
+-- Safe to run multiple times: UPDATE targets only the legacy value.
+UPDATE project_meta
+SET value = 'BAAI/bge-m3'
+WHERE key = 'embedder_name'
+  AND value = 'bge-m3-Q8_0.gguf';


petersimmons1972:
Superseded by #972 (merged) — issue #911 fixed there. Closing this duplicate draft.

PR #959, opened 2026-06-01
search: cache embedder metadata check

### LINKED ISSUE #929: hygiene(embed): checkEmbedderMeta() makes 2 DB round-trips on every recall — add process-scoped cache
## Summary

`checkEmbedderMeta()` is called on every `RecallWithOpts()`, `RecallWithinMemory()`, `Store()`, and `StoreBatch()` call. It issues `backend.GetMeta()` twice: once for `embedder_name` and once for `embedder_dimensions`. The embedder name almost never changes between calls within a process lifetime.

## Impact

At 100 recalls/second (busy deployment), this is 200 extra DB queries per second that are no-ops in the steady state. Under a 25-connection pool these are cheap but non-zero.

## Fix

Add a `sync.Once` or atomic boolean `embedderMetaVerified` on the SearchEngine that is set after the first successful check. Reset it only in `MigrateEmbedder`. Skip the DB reads after verification.

## Personas

skeptical-staff-engineer

## References

From QA sweep #908 (round 1).
`internal/search/engine.go` line 1036 (`checkEmbedderMeta`).

petersimmons1972:
## Summary
- Added one-time, process-scoped caching in  so repeated calls to recall/store paths do not re-run  DB reads after first success.
- Added locking around cache state to avoid concurrent duplicate checks.
- Reset the cache in  so metadata verification is redone after embedder migration.

Commit: 

Closes #929

### DIFF
--- a/docs/proposed-fixes/2026-06-01-engram-go-branch-reconciliation.md
+++ b/docs/proposed-fixes/2026-06-01-engram-go-branch-reconciliation.md
@@ -1,0 +1,32 @@
+# Engram-Go Branch Reconciliation Triage (2026-05-31 snapshot)
+
+Context: this note records the decision for stale and rescued branches identified in
+[`#942`](https://github.com/petersimmons1972/engram-go/issues/942), aligned with the
+reconciliation snapshot taken 2026-05-31.
+
+## Action decisions
+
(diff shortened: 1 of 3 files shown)


petersimmons1972:
Superseded by #972 (merged) — issue #929 fixed there. Closing this duplicate draft.

Issue #938, opened 2026-05-31
LME score recovery: prioritize retrieval recall lift toward 90%


petersimmons1972:
## Summary
Engram-go LME benchmark is showing ~35% exact score (including partially-correct ~45%) after scoring 500 rows. Primary failure mode appears retrieval:

- ~168/271 incorrect rows have gold session missing from top-10 recall.
- ~80/271 incorrect rows have gold in top-10 but wrong answer, suggesting noisy ranking/context packing issues.
- ~5/271 correct/partial cases lack retrieval metrics.

The benchmark content and answers are fixed; only retrieval/config can change.

## Current baseline
- Existing rows only (no re-ingest required for this pass).
- Existing benchmark harness supports repair/re-score flow and multi-model sequencing.

## Proposed improvements
1. Add retrieval candidate fusion (vector + BM25/FTS + query variants + lexical/exact boosts) to improve gold session recall.
2. Make session-anchor and session-level evidence a first-class retrieval source (not only post-recall expansion).
3. Add exact-fact/identifier scoring boosts for entities, URLs, phone numbers, song/venue names.
4. Improve question rewriting/query variants for LME-style prompts.
5. Improve context packing and evidence-first ordering to reduce wrong answers from noisy top hits.
6. Add retrieval-only audit + recall gates before full benchmark runs.

## Acceptance target
-  should increase enough to support overall  LME score on benchmark without changing questions/answers.
- Keep changes behind config flags for safe rollback.

## Notes
- Prior work already includes  plumbing in search/recall path and some query/anchor handling.
- This issue is for targeted follow-on implementation and validation.,

petersimmons1972:
## Clarification / Implementation Notes

- Baseline: ~35% exact score on 500 rows (incl. partials ~45%).
- Retrieval diagnostics from current dataset: 168 incorrect rows had gold session missing from top-10, 80 had gold in top-10 but wrong answer, ~23 had missing metrics.
- This confirms recall-first fixes are higher leverage than model changes.

Suggested rollout:
1) Retrieval-only audit gate that computes gold in top1/top5/top10/top50 + gold chunk in context.
2) Candidate fusion across vector + BM25/FTS + query variants + identifier/lexical boosts.
3) Session-first retrieval path before final ranking.
4) Evidence-first context packing and post-retrieval filtering.
5) Config-flag gated rollout + staged benchmarking.

[...]

PR #983, opened 2026-06-01
feat(lme): preference-anchoring generation prompt (#938)

petersimmons1972:
## Summary

Rank-2 generative lever targeting the 23/27 generative failures where the gold preference session is in-context but the generator averages across sessions instead of anchoring on it.

- Adds `--preference-anchor` flag (default **off**) to `cmd/longmemeval` — gated on `QuestionType == "single-session-preference"`, no-op for all other types.
- Injects three explicit anchoring instructions into the preference generation prompt: identify the single most-relevant session, do not average across sessions, specific product/platform/style details override general inferred preferences.
- Change is localised to `internal/longmemeval/claude_additions.go` (new `GenerationPromptForTypeWithPreferenceAnchor`) and the prompt-assembly switch in `cmd/longmemeval/run.go`. Retrieval/ordering code is not touched.
- 3 new unit tests (on / off / no-op for other types). All existing tests pass.

**HELD pending eval**: anchor+H15 vs H15-alone on the 30-item preference subset before merge.

## Test plan

- [ ] `go build ./...` — clean
- [ ] `go vet ./cmd/longmemeval/...` — clean
- [ ] `go test ./internal/longmemeval/ -run TestGenerationPromptForTypeWithPreferenceAnchor -count=1` — 3 PASS
- [ ] `go test ./cmd/longmemeval/ -count=1` — full suite passes
- [ ] `go test ./internal/longmemeval/ -count=1` — full suite passes
- [ ] Eval agent: run with `--preference-anchor --dual-preference-recall` vs `--dual-preference-recall` alone on 30-item preference subset


petersimmons1972:
Closing — falsified by Tier-1 eval. +H15+anchor scored 1/30 CORRECT vs +H15-alone 3/30 (−2 CORRECT): forcing the generator to commit to 'the single most relevant session' made it pick the WRONG session (most BM25-similar ≠ gold, with ~480 sessions/item), erasing H15's gains. Root-cause diagnostic: 48% of preference failures are retrieval-precision (gold in top-100 but not the top-15 context), not generation — so the next lever is a top-100 re-rank (preference-signal boost) to get the gold session into context. A softer generation lever may be revisited for the 37% generation-miss (C-class) AFTER precision is fixed. Branch left intact; learning persisted to Engram.

PR #952, opened 2026-06-01
feat: retrieval fusion + exact-signal packing gates for LME recovery (#938)

petersimmons1972:
## Summary
- add optional retrieval fusion in `longmemeval run` by unioning primary/raw/exact-signal query variants
- add optional exact-signal ranking boost for candidate context IDs
- add optional evidence-first context packing before prompt assembly
- cover the new behavior with focused unit tests

## Flags (all default off)
- `--retrieval-fusion`
- `--exact-signal-boost`
- `--evidence-first-pack`

## Validation
- `go test ./cmd/longmemeval`
- `go test ./internal/longmemeval`

Closes #938

PR #984, opened 2026-06-01
feat(lme): preference-signal precision re-rank (#938)

petersimmons1972:
## Summary

- Adds `--preference-rerank` flag (default OFF, gated) to `cmd/longmemeval`
- Targets the 48% class-B precision miss: gold session recalled in top-100 but truncated before the top-15 context window
- Before truncating top-100→context-topk for `single-session-preference` questions, re-ranks candidates by: `blended = 0.5 * normalized_vector_score + 0.5 * pref_signal_score` where pref_signal_score is TF-overlap against question content terms + preference-signal tokens
- Insertion point: `run.go:preferenceRerankIDs()` called from `runOne()` just before `selectContextIDs()` — the exact truncation-to-topk boundary
- Unit tests verify scoring correctness and that a gold at rank 20 lifts into top-15 after re-rank

## HELD pending eval

Held pending the +H15+rerank eval arm. Merge decision: ≥+2 CORRECT delta with ≤+1 INCORRECT regression.

## Test plan

- [x] go build clean
- [x] go vet clean
- [x] go test ./cmd/longmemeval/... -count=1 -race all pass
- [ ] Eval arm result comparison vs H15-alone (3/16/11)
- [ ] B-class recovery check on ~13 known class-B items

### DIFF
--- a/cmd/longmemeval/main.go
+++ b/cmd/longmemeval/main.go
@@ -81,6 +81,12 @@ type Config struct {
 	ExactSignalBoost    bool // re-rank candidate IDs by exact identifier/entity overlap
 	EvidenceFirstPacked bool // order context blocks by exact overlap before prompt assembly
 
+	// #938 precision re-rank: before truncating top-100→context-topk for
+	// single-session-preference questions, re-rank the top-100 candidates by a
+	// blended score that boosts candidates containing preference-signal tokens
+	// and question content terms. Default OFF; enabled by --preference-rerank.
+	PreferenceRerank bool
+
 	// #749: contention guard
 	ExclusiveBackend bool   // guard the vLLM endpoint with a PID-liveness lockfile (default true)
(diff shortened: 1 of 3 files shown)


petersimmons1972:
Closing — falsified by eval. +H15+precision-rerank scored 1/30 CORRECT vs H15-alone 3/30 (−2). The BM25 preference-signal (prefer/like/enjoy/usually) is too broad: it lifted false-positive sessions into the top-15, displacing context H15 had ranked correctly (2 CORRECT→PARTIAL). Only 2/13 class-B gold sessions recovered (both INCORRECT→PARTIAL). Meta-lesson: within the raw-session representation, no crude signal — BM25 overlap here, generation anchoring in #983, retrieval breadth in #952 — reliably separates 'the session where the user stated the preference' from topically-similar distractors. Architectural limit, not a tuning gap; reinforces the extraction-first clean-sheet direction. Branch left intact.

Issue #989, opened 2026-06-02
chore(recall): generalize hardcoded "embed_timeout" reason in handle-mode WARN log + MCP response


petersimmons1972:
## Background

Flagged as a cosmetic nit during the 3-round review of #973/#917 (merged in PR #988). Not a blocker — metric accuracy is unaffected because the `RecallDegradedTotal` Prometheus counter uses the correct `reason` label at increment time.

## Problem

In the `handle` mode path, the degradation reason string is hardcoded as `"embed_timeout"` in two places:

1. The WARN-level structured log field (`reason="embed_timeout"`)
2. The MCP response body (`"degradation_reason": "embed_timeout"`)

This means that if a future degradation cause is added (e.g. `"embed_pool_exhausted"`, `"embed_model_unavailable"`), the log and MCP response will still report `"embed_timeout"` even if the actual reason differs. The counter label is correct; only these two display surfaces are hardcoded.

## Fix

Pass the actual reason string through the call chain to the `handle` mode formatter, replacing the hardcoded literal with the dynamic value already available at the engine layer.

## Scope

Cosmetic / observability quality. No behavior change, no metric change, no breaking change. Safe to batch with other small cleanups.

petersimmons1972:
Fixed in merged PR #1006 (tool_timeout rename)

Issue #973, opened 2026-06-01
fix(cli): --embed-recall-timeout-ms=0 does not mean 'no timeout' as documented


petersimmons1972:
## Summary

The `--embed-recall-timeout-ms` flag help text says `0 = no timeout`, but passing `0` does not disable the timeout — it silently falls back to the 500ms default.

## Root cause

`main.go` calls `engine.SetEmbedRecallTimeout(ms)` after construction. `SetEmbedRecallTimeout` only applies when `ms > 0`:

```go
func (e *SearchEngine) SetEmbedRecallTimeout(ms int) {
    if ms > 0 {
        e.embedRecallTimeout = time.Duration(ms) * time.Millisecond
    }
    // ms == 0 → no-op; embedRecallTimeout stays at the env/default value
}
```

Separately, `RecallWithOpts` also has a `if embedTimeout <= 0` guard that falls back to the 500ms default — but this is never reached when passing `0` via the flag, because `SetEmbedRecallTimeout(0)` is already a no-op and the field was initialized to a positive value by `New()`.

## Actual behavior when `--embed-recall-timeout-ms=0`

The flag call is a no-op. The engine uses whatever `ENGRAM_EMBED_RECALL_TIMEOUT_MS` env var says (defaulting to 500ms). The timeout is never disabled.

## Expected behavior (per documented contract)

Passing `0` should disable the embed timeout entirely, allowing the parent context's deadline to govern the embed call.

## Fix scope (not in this issue)

To honor `0 = no timeout`:
1. Change `SetEmbedRecallTimeout` to accept `ms == 0` as a valid override (sentinel for "unbounded")
2. Remove the `if embedTimeout <= 0` fallback guard in `RecallWithOpts` (or handle it as "no timeout" instead of defaulting to 500ms)
3. Update the `RecallWithinMemory` path similarly

**Do NOT fix in PR #971** — that PR restores the env-var regression only.

petersimmons1972:
Fixed in merged PR #978 (noEmbedTimeout sentinel)

PR #998, opened 2026-06-02
chore(recall): generalize hardcoded embed_timeout reason in handle-mode WARN + MCP response (#989)

petersimmons1972:
## Summary

- **#973** — Already satisfied by PR #988 (merged 2026-06-02). `SetEmbedRecallTimeout(0)` on `origin/main` correctly maps `0` to the `noEmbedTimeout` sentinel via the `else if ms == 0` branch added in #988. The CLI flag path (`main.go:380` calls `engine.SetEmbedRecallTimeout(*embedRecallTimeoutMS)`) passes `0` through unchanged, which is now handled correctly. No code change required.
- **#989** — Genuinely open. Three places in `tools_recall.go` hardcoded `"embed_timeout"` as the degradation reason in the handle-mode WARN log and MCP response body, even though the engine already computed the correct reason (`"embed_timeout"` vs `"embed_error"`). Fixed by threading the reason through `RecallOpts.EmbedDegradedReason`.

## Changes

- `internal/search/engine.go`: Add `EmbedDegradedReason *string` to `RecallOpts`; populate it in `RecallWithOpts` alongside the existing `EmbedDegraded` bool
- `internal/mcp/tools_recall.go`: Capture `EmbedDegradedReason` from opts; use the dynamic value in handle-mode `degradedMap`, `addRecallDegradedWarning`, and the non-handle-mode fallback (replaces 3 hardcoded `"embed_timeout"` literals)
- `internal/search/embed_deadline_test.go`: 3 new tests for `EmbedDegradedReason` (timeout reason, hard-error reason, no-degrade stays empty)
- `internal/mcp/degradation_test.go`: 2 new integration tests verifying the handle-mode response carries the dynamic reason

## Verification

- `go build -buildvcs=false ./...` — clean
- `go test -buildvcs=false ./cmd/engram/... ./internal/search/... ./internal/mcp/... -count=1 -race` — all pass
- `golangci-lint run ./internal/search/... ./internal/mcp/... ./cmd/engram/...` — 0 issues

## Review

3-round adversarial review required per AI-PR policy.

### DIFF
--- a/internal/mcp/degradation_test.go
+++ b/internal/mcp/degradation_test.go
@@ -2,12 +2,14 @@ package mcp
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"testing"
 
 	mcpgo "github.com/mark3labs/mcp-go/mcp"
 	"github.com/petersimmons1972/engram/internal/claude"
 	"github.com/petersimmons1972/engram/internal/db"
+	"github.com/petersimmons1972/engram/internal/embed"
 	"github.com/petersimmons1972/engram/internal/search"
 	"github.com/petersimmons1972/engram/internal/types"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


petersimmons1972 (commented):
## AI-PR-Policy Review — Round 1/2/3 Combined
### PR #998 · `chore(recall): generalize hardcoded embed_timeout reason (#989)`

---

### ROUND 1: CORRECTNESS

**Finding C-1 — MEDIUM (nice-to-have): `structuredEmbedDegradedError` path discards the reason silently.**

Location: `internal/mcp/tools_recall.go`, lines ~416-417 (the `DegradedErrorMode == "structured"` early-return).

The engine correctly sets `embedDegradeReason` before the MCP layer reads it. However, the `structuredEmbedDegradedError` early-return fires *after* the engine populates the opts pointer but *before* any response field carries the reason. The structured-error response body contains `"code": "embed_pipeline_degraded"` but no `reason` field. A consumer receiving this response cannot distinguish timeout from hard-error. This was equally true before the patch (the reason was never in that branch), so this PR does not regress it — but it does not fix it either, and the fix is now trivially available (`embedDegradeReason` is in scope at that call site). Not a blocker; the structured-error path is opt-in and rare. File a follow-up.

**Finding C-2 — PASS: reason accuracy in all branches.**

- Timeout path: `embedCtx.Err() != nil` → `"embed_timeout"`. Correct: the per-embed deadline fired, which is the bounded sub-context, not the caller's context.
- Hard-error path: any other non-nil `err` → `"embed_error"`. Correct.
- No-degrade path: `embedDegradeReason` stays `""`, opts pointer receives `""`. Correct.
- Health-check circuit-open path (non-handle branch): `cfg.EmbedderHealth.Snapshot()` returns its own reason string and this PR does not touch that path. The `embedDegradeReason` is only used as fallback when `reason == "" && embedDegraded`. Circuit-open would have its own non-empty reason from `Snapshot()`, so the fallback line is never reached in that case. Correct.

**Finding C-3 — PASS: nil-pointer safety.**

Both `opts.EmbedDegraded` and `opts.EmbedDegradedReason` are only written when `!= nil`. The MCP layer declares concrete `bool` and `string` locals and takes their addresses, so the pointers are always non-nil on the call path reviewed. No nil-deref is possible.

**Finding C-4 — PASS: no data race on the shared RecallOpts pointer.**

All three `RecallWithOpts` call sites in `handleMemoryRecall` are inside a single if/else-if/else branch — exactly one executes per request. `opts` is stack-allocated in the handler. The engine writes back through the pointer sequentially before returning. No goroutine shares the opts struct. There is no race condition.

**Finding C-5 — PASS: old hardcoded `"embed_timeout"` was wrong, new dynamic value is correct.**

The original hardcoded `"embed_timeout"` was incorrect whenever the embedder failed with a hard error (connection refused, model crash). The new value is derived from the same `degradeReason` variable already used by the metrics label, making the MCP-surface reason consistent with the prometheus counter label. That is the right source of truth.

**Finding C-6 — MEDIUM (nice-to-have): handle-mode `addRecallDegradedWarning` called with empty reason when `embedDegradeReason == ""`.**

Location: `tools_recall.go` handle-mode branch, line ~429.

`addRecallDegradedWarning(out, cfg.RouterURL, embedDegradeReason)` is only called when `embedDegraded == true`. If `embedDegraded` is true but `embedDegradeReason` is somehow empty (not possible via the current engine path, but conceivable if the interface contract drifts), the warning log would carry `reason=""`. `degradedMap` already guards this (it omits the key when reason is empty), but `addRecallDegradedWarning` passes the empty string to `slog.Warn` without a guard. Low-blast-radius; the log is malformed but harmless. Not a blocker.

---

### ROUND 2: COVERAGE

**Finding CV-1 — PASS: search-layer timeout/hard-error/no-degrade branches covered.**

`TestEmbedDegradedReason_Timeout` uses a blocking embedder against the 500ms default deadline — this exercises the `embedCtx.Err() != nil` branch and asserts `reason == "embed_timeout"`. Solid.

`TestEmbedDegradedReason_HardError` uses an immediate-return error embedder — this exercises the else branch and asserts `reason == "embed_error"`. The assertion is exact, not just non-empty.

`TestEmbedDegradedReason_NoDegrade` asserts both `degraded == false` and `reason == ""`. This is the correct negative case.

**Finding CV-2 — PASS: MCP-layer tests cover the actual JSON response body.**

`TestHandleMode_DegradedReason_IsEmbedError` unmarshals the response body and asserts `degraded.reason == "embed_error"`. This is a real end-to-end assertion through the handler, not just a unit check on the engine.

`TestHandleMode_DegradedReason_Nil_WhenNoDegrade` confirms the reason key is absent when not degraded — which tests `degradedMap`'s conditional key emission. Correct.

**Finding CV-3 — MEDIUM (nice-to-have): no MCP-layer test for the non-handle (full-results) branch with `EmbedderHealth.Snapshot()` returning a non-empty reason vs. falling through to `embedDegradeReason`.**

The non-handle path has a three-way interaction: `Snapshot()` health-check reason, `embedDegraded` flag, and the new `embedDegradeReason`. The existing `embed_degraded_error_test.go` presumably covers the non-handle path in general, but there is no test in this PR that specifically asserts: "when `Snapshot()` returns `reason == ""` and `embedDegraded == true`, the response carries the engine's reason rather than an empty string." The fix at line ~452 is correct but untested at the MCP layer. Not a blocker given the logic is simple (one condition), but a gap worth noting.

**Finding CV-4 — PASS: `immediateErrEmbedder` in degradation_test.go implements the full `embed.Client` interface.**

The compile-time guard `var _ embed.Client = immediateErrEmbedder{}` confirms interface satisfaction. No silent stub hole.

---

### ROUND 3: STRUCTURAL

**Finding S-1 — MEDIUM (nice-to-have): `*string` for reason vs. a typed constant.**

The doc comment enumerates exactly two valid values: `"embed_timeout"` and `"embed_error"`. These are already used as prometheus label values via `metrics.RecallDegradedTotal.WithLabelValues(degradeReason)`. Using raw strings means a future third cause (e.g., circuit-open propagated through the engine) can silently introduce a value not covered by any test or validation. A `DegradeReason` type alias with named constants would close this. This is a design preference rather than a correctness bug, and the values are already constrained by the metrics label surface. Nice-to-have; not a blocker.

**Finding S-2 — PASS: `RecallOpts` mutation pattern is consistent with existing convention.**

`EmbedDegraded *bool` was already an output field on `RecallOpts` before this PR. `EmbedDegradedReason *string` follows the identical pattern. The engine writes back through the pointer only when non-nil, which is the established convention. This is not a new side-effect smell; it is an extension of the existing out-parameter idiom already present and apparently accepted in this codebase.

**Finding S-3 — MEDIUM (nice-to-have): naming inconsistency between field and local variable.**

The struct field is `EmbedDegradedReason` (past-participial, noun phrase). The local variable in both the engine and the MCP layer is `embedDegradeReason` (verb form). Minor, but it introduces a read friction when navigating: a reader scanning for the field name will not immediately match the local. Trivial to fix in a follow-up; does not affect correctness.

**Finding S-4 — PASS: placement in `RecallOpts` struct is correct.**

New field placed immediately after `EmbedDegraded`, which is its semantic companion. Grouped correctly. Doc comment is accurate and documents both valid values and the nil-means-skip contract.

**Finding S-5 — PASS: no unnecessary complexity introduced.**

The engine change is four lines: a local string initialisation, assignment inside the error branch, and a guarded pointer write. The MCP change replaces three hardcoded string literals with the captured local. Subtractive in nature — this removes the incorrect literal and wires in the correct source. Nothing left to remove from the change itself.

---

### VERDICT

**SHIP.**

Zero severity/blocker findings. Five medium nice-to-haves identified:

| ID | Location | Summary |
|----|----------|---------|
| C-1 | `tools_recall.go` ~416 | `structuredEmbedDegradedError` response omits reason; fix now trivial |
| C-6 | `tools_recall.go` ~429 | handle-mode warning log passes empty reason without guard |
| CV-3 | `degradation_test.go` | No MCP test for non-handle path's `Snapshot()+embedDegradeReason` fallback |
| S-1 | `engine.go` `RecallOpts` | `*string` vs typed constant for reason values |
| S-3 | Both files | `EmbedDegradedReason` (field) vs `embedDegradeReason` (local) naming inconsistency |

The core correctness of the change is sound: reason values are accurate, nil-safety is maintained, no race condition exists, the RecallOpts mutation pattern is consistent with the established convention, and the five tests provide meaningful assertions on exact reason values at both layers. Recommend tracking C-1 as a follow-up issue given the fix is now a one-liner.

PR #1061, opened 2026-06-05
[codex-poll recovered] petersimmons1972/engram-go#1060

### LINKED ISSUE #1060: reembed worker-pool: remove per-batch barrier for additive multi-card embeds
Filed by queue-agent on 2026-06-05T12:39:51Z at the request of: Peter Simmons

## Context

The `engram-reembed` Rust worker (`reembed-rs/src/main.rs`) cannot benefit from multiple embed backends of unequal speed. Measured on the live ~1.9M-chunk snowflake→bge-m3 reembed (2026-06-05): load-balancing across leviathan (RX 7900 XT, gfx1100) + precision W6800 (gfx1030) ran at 1,533/min vs 3,789/min on leviathan alone — a 57% regression — because the worker blocks on the slowest backend every batch cycle.

Root cause: `run_batch` claims `batch_size` (2000) rows, fans them into `embed_sub_batch` (100) chunks dispatched under one shared `Semaphore`, then barriers on `for h in handles { h.await }` before the main loop claims the next batch. The slowest sub-batch (routed to the slow card by olla) gates each cycle, the shared semaphore lets a slow request starve the fast card's concurrency, and the batch-latency-based `AdaptiveConcurrency` ramps down when one backend is slow. The fix is to replace the claim→fan-out→barrier model with a continuous worker pool: N persistent workers each independently claim a small `SKIP LOCKED` slice, embed, update, and loop — no barrier — so a fast backend simply completes more cycles and adding a card is purely additive.

## Files

- `reembed-rs/src/main.rs` — replace `run_batch` + the `loop` in `run` with a continuous worker-pool model: spawn N worker tasks, each looping {claim `embed_sub_batch` rows via `FOR UPDATE SKIP LOCKED` in its own short tx, embed via olla, `UPDATE chunks SET embedding`}; remove the per-batch `await`-all barrier and the shared `Semaphore`; replace batch-latency `AdaptiveConcurrency` with a fixed worker count (`ENGRAM_REEMBED_CONCURRENCY_MAX`, default preserved) or per-request-latency adaptation; preserve startup probe, connection-pool warmup, graceful ctrl_c shutdown (drain in-flight workers), and structured logging (per-worker throughput + aggregate gauge).
- `reembed-rs/src/claim.rs` — (new) extract the claim query + per-slice embed→update unit into a pure-ish function `process_slice(pool, http, cfg) -> Result<(usize claimed, usize failed)>` so it is unit-testable independent of the worker loop.
- `reembed-rs/tests/worker_pool.rs` — (new) integration tests: SKIP LOCKED claim correctness under concurrent workers (no double-processing), slow-backend isolation (a slow mock embed client does not reduce a fast worker's completion count), graceful shutdown drains in-flight work.
- `reembed-rs/Cargo.toml` — add dev-dependencies for testing (e.g. `wiremock` or a mock HTTP client for the embed endpoint; `sqlx` test features / `testcontainers` for an ephemeral Postgres). Keep runtime deps unchanged.
- `reembed-rs/README.md` (or inline module doc) — document the worker-pool model and the `ENGRAM_REEMBED_*` env knobs that change meaning (concurrency now = worker count; `BATCH_SIZE` retired or repurposed as per-claim slice size alias of `EMBED_SUB_BATCH`).

## Acceptance (TDD)

- [ ] Concurrent `SKIP LOCKED` claims never double-process a chunk (two workers claim disjoint row sets).
  - Test: `reembed-rs/tests/worker_pool.rs::skip_locked_no_double_processing`
- [ ] A slow embed backend does not reduce a fast worker's throughput (additive, not straggler-gated): with one fast + one slow mock client, total completed ≥ fast-only completed within a fixed wall-clock window.
  - Test: `reembed-rs/tests/worker_pool.rs::slow_backend_does_not_gate_fast`
- [ ] `process_slice` claims at most `embed_sub_batch` rows, embeds them, writes embeddings, and reports failed count without panicking on embed error/timeout.
  - Test: `reembed-rs/tests/worker_pool.rs::process_slice_happy_and_failure_paths`
- [ ] Graceful shutdown: on ctrl_c, in-flight slices finish and their embeddings are persisted before exit (no torn writes, no orphaned NULLs that were claimed).
  - Test: `reembed-rs/tests/worker_pool.rs::graceful_shutdown_drains_inflight`
- [ ] Startup embed probe + connection-pool warmup behavior preserved (worker does not start claiming until probe succeeds; exits non-zero after bounded retries if endpoint unreachable).
  - Test: `reembed-rs/tests/worker_pool.rs::startup_probe_gates_workers`

## Spec refs

- GitHub issue petersimmons1972/engram-go#1059 — reembed worker: replace per-batch barrier with continuous worker pool
- Current implementation: `reembed-rs/src/main.rs` (`run_batch` lines ~141-292, main `loop` lines ~429-475, `AdaptiveConcurrency`, startup probe lines ~314-406)
- petersimmons1972/aifleet#399 (fleet authoritative embed model + two-card duty split), #401 (post-reembed duty model), #385 (provider/consumer contract), #400 (FC-sourced endpoints lack `embeddings` capability), #95 (per-endpoint LB weights)
- Measured straggler evidence table — issue #1059 body

## Constraints

- Must not double-process chunks: every claim path uses `FOR UPDATE SKIP LOCKED` in a short transaction committed before the embed call (preserve the existing claim-then-release pattern).
- Must preserve `ORDER BY id DESC` claim ordering (newest chunks become searchable first, #914) and the partial-batch/backoff sleep behavior when the queue is empty.
- Must preserve the startup embed probe (bounded retry, exit-1 on failure) and the connection-pool warmup / `pool_max_idle_per_host` tuning.
- Must preserve graceful shutdown on ctrl_c (no abrupt drop of in-flight slices).
- Embed routing stays through olla (`LITELLM_URL`); the worker MUST NOT address backends directly or know which card it hits — speed-proportionality must emerge from the pool dynamics, not hardcoded host knowledge.
- Hardware safety: this change must NOT re-introduce load on the precision Radeon VII (gfx906); embed-backend selection is owned by olla/fleet config (#399/#401), not the worker.
- TDD: failing test before implementation; every commit includes a test. Must pass existing build/clippy with no regressions.
- Branch `agent/codex/issue-1059-reembed-worker-pool`; draft PR; do not push to main/master. Open PR then register auto-merge (`gh pr merge --auto --squash --delete-branch`); confirm rebased on origin/main and CI green.

## Out of scope

- Per-endpoint load-balancer weights in olla — tracked in aifleet#95.
- FC-sourced endpoints lacking the `embeddings` capability — tracked in aifleet#400.
- Fleet authority / two-card duty split config — tracked in aifleet#399, #401.
- Container build + deploy of the new image (manual, founder-gated; not part of the code PR).
- Thermal-aware fleet (GPUHost CRD temp reporting / Prometheus exporter) — separate work.

petersimmons1972:
Auto-recovered unpublished Codex work for #1060 (claude-codex#22). Draft — needs review/CI. Closes #1060

### DIFF
--- a/reembed-rs/Cargo.lock
+++ b/reembed-rs/Cargo.lock
@@ -24,6 +24,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "atoi"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -179,6 +189,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "deadpool"
+version = "0.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0be2b1d1d6ec8d846f05e137292d0b89133caf95ef33695424c09568bdd39b1b"
+dependencies = [
+ "deadpool-runtime",
+ "lazy_static",
+ "num_cpus",
+ "tokio",
+]
+
+[[package]]
+name = "deadpool-runtime"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "092966b41edc516079bdf31ec78a2e0588d1d0c08f78b91d8307215928642b2b"
+
+[[package]]
 name = "der"
 version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -241,6 +269,7 @@ dependencies = [
  "tokio-util",
  "tracing",
  "tracing-subscriber",
+ "wiremock",
 ]
 
 [[package]]
@@ -299,6 +328,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "fnv"
+version = "1.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
 name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -311,6 +346,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cb4cb245038516f5f85277875cdaa4f7d2c9a0fa0468de06ed190163b1581fcf"
 dependencies = [
  "percent-encoding",
+]
+
+[[package]]
+name = "futures"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b147ee9d1f6d097cef9ce628cd2ee62288d963e16fb287bd9286455b241382d"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-executor",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
 ]
 
 [[package]]
@@ -358,6 +408,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cecba35d7ad927e23624b22ad55235f2239cfa44fd10428eecbeba6d6a717718"
 
 [[package]]
+name = "futures-macro"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e835b70203e41293343137df5c0664546da5745f82ec9b84d40be8336958447b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "futures-sink"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -375,8 +436,10 @@ version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "389ca41296e6190b48053de0321d02a77f32f8a5d2461dd38762c0593805c6d6"
 dependencies = [
+ "futures-channel",
  "futures-core",
  "futures-io",
+ "futures-macro",
  "futures-sink",
  "futures-task",
  "memchr",
@@ -422,6 +485,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "h2"
+version = "0.4.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "171fefbc92fe4a4de27e0698d6a5b392d6a0e333506bc49133760b3bcf948733"
+dependencies = [
+ "atomic-waker",
+ "bytes",
+ "fnv",
+ "futures-core",
+ "futures-sink",
+ "http",
+ "indexmap",
+ "slab",
+ "tokio",
+ "tokio-util",
+ "tracing",
+]
+
+[[package]]
 name = "hashbrown"
 version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


petersimmons1972:
## Coordinator review — REQUEST CHANGES (2 independent reviews: zero-context + adversarial)

Core logic is **sound**: no data loss (rows stay NULL on embed failure → retried; double-writes are idempotent), no barrier (straggler fix achieved), shutdown drains in-flight, startup probe preserved, routes via olla only. But the tests don't actually prove this PR's central guarantee, plus one observability bug. Fix before merge:

### BLOCKING

1. **`skip_locked_no_double_processing` uses `workers=1` — it cannot detect the very race it's named for.** The claim tx commits *before* the embed (lock released while `embedding IS NULL`), so two concurrent workers CAN re-claim the same rows in the `[claim-commit, UPDATE]` window. Result is idempotent (same text→same vector) but it's wasted work and the test gives false assurance.
   - **Fix:** make it `workers≥4, slice=2, rows=20`, assert dupe-update count == 0 via the `updated_count` trigger. Add a code comment in `claim.rs` documenting the commit-before-embed window + idempotency safety net (the old `main.rs` had this comment; it was dropped).

2. **`slow_backend_does_not_gate_fast` assertion is too weak.** `fast_mixed + slow_mixed >= fast_baseline` passes even if the slow worker fully gates the fast one. This PR's whole purpose is additive non-gating — the test must prove it.
   - **Fix:** assert the fast worker *individually* achieves ≥ ~80% of its solo baseline.

3. **`completed_total` gauge counts failed/timed-out slices.** `process_slice` returns `(len, len)` on failure and `run_worker` adds `claimed` to `completed`, so logs show `completed_total` climbing during 100%-failure. Misleading production observability.
   - **Fix:** separate "attempted" vs "written"; only count successfully-written rows in `completed_total`.

### SHOULD-FIX

4. `graceful_shutdown_drains_inflight` asserts only `done > 0` — strengthen to assert all in-flight claimed rows persisted.
5. **`AdaptiveConcurrency` + its env vars (`*_CONCURRENCY_MIN/MAX`, `*_LATENCY_*`) are now dead runtime code** (fixed worker count). Remove, or clearly mark reserved. ⚠️ Note: this removes the old latency-based auto-throttle — a thermal/latency-aware governor will be added as a separate follow-up (founder request).
6. **`ENGRAM_REEMBED_BATCH_SIZE` default changed 100→8** — silent ~12× throughput drop for deployments relying on the old default. Add a deploy-runbook note. (Our live deployment sets it explicitly, so unaffected.)
7. Empty-queue backoff ceiling 300s is high (rows can wait up to 5 min after refill); lower to ~30-60s.
8. `ordering_tests` asserts a hardcoded query string, not the actual query in `claim.rs` — update to reference the real query so it can catch regressions.

CI is green (compiles/clippy/tests pass without a DB), so these are logic/coverage issues CI won't catch. Re-request review after fixes.

petersimmons1972:
Addressed the coordinator blocking items in appended commit 927449f (no force-push).\n\nChanges:\n- Strengthened skip_locked_no_double_processing to workers=4, slice=2, rows=20 and assert updated_count duplicate writes stay at 0.\n- Added the claim.rs comment documenting the commit-before-embed window and conditional-update idempotency safety net.\n- Strengthened slow_backend_does_not_gate_fast so the fast worker individually reaches at least 80% of solo baseline.\n- Split reembed slice accounting into attempted/written/failed; completed_total now increments only by written rows, with attempted_total logged separately.\n- Also strengthened graceful_shutdown_drains_inflight to prove the in-flight 4-row slice persists, fixed the Rust runner JoinHandle compile issue, serialized env-mutating config tests, and locked the Rust test deps in Cargo.lock.\n\nLocal verification:\n- TEST_DATABASE_URL=postgres://...@localhost:55432/engram_test?sslmode=disable cargo test --test worker_pool skip_locked_no_double_processing -- --nocapture: pass\n- TEST_DATABASE_URL=postgres://...@localhost:55432/engram_test?sslmode=disable cargo test --test worker_pool process_slice_happy_and_failure_paths -- --nocapture: pass\n- TEST_DATABASE_URL=postgres://...@localhost:55432/engram_test?sslmode=disable cargo test --test worker_pool slow_backend_does_not_gate_fast -- --nocapture: pass\n- TEST_DATABASE_URL=postgres://...@localhost:55432/engram_test?sslmode=disable cargo test --test worker_pool graceful_shutdown_drains_inflight -- --nocapture: pass\n- TEST_DATABASE_URL=postgres://...@localhost:55432/engram_test?sslmode=disable cargo test: pass\n- go test ./... -count=1: pass\n- git diff --check: clean\n\nCurrent PR state: CI is running on 927449f; not merging while draft/review/CI gates remain open.

Issue #1087, opened 2026-06-13
exec-contract(reembed): first hardening lane for cross-drainer safety + Rust TDD fixes


petersimmons1972:
## Purpose

Convert #1085 from broad synthesis into a **tight execution contract** for the first implementation lane. This issue is intentionally narrow: first ship the smallest changes that restore a universal no-overwrite invariant and eliminate the live Rust dead-backend hot-claim loop.

Parent/context:
- supersedes discussion-heavy execution planning in #1085 for the first implementation slice
- does **not** reopen architecture: worker-pool + `SKIP LOCKED` remains the baseline

## Why this lane exists

#1085 now has good diagnosis but too much sequencing drift. This issue defines the first code-and-tests lane with explicit order, files, RED/GREEN gates, and deferrals.

### Patch A — universal idempotent write guard (correctness first)

**Goal:** make the claimed invariant true across *all* drainers, not only `reembed-rs`.

**Problem evidence:**
- Rust path already guards writes:
  - `reembed-rs/src/claim.rs`
  - `UPDATE chunks SET embedding = $1::vector WHERE id = $2 AND embedding IS NULL`
- Go paths currently do not:
  - `internal/reembed/global_worker.go`
  - `internal/embedgateway/drain.go`
  - `internal/db/postgres_chunk.go` (`UpdateChunkEmbedding`)

Until all write paths share the same invariant, mixed-drainer races are not merely "wasted compute"; they can overwrite another drainer's write.

**Files:**
- `internal/reembed/global_worker.go`
- `internal/embedgateway/drain.go`
- `internal/db/postgres_chunk.go`
- tests in the owning packages

**Required code change:**
Use `AND embedding IS NULL` on all chunk-embedding UPDATEs.

**Acceptance (TDD):**
Write failing tests first proving:
1. second writer gets `rows_affected == 0` instead of overwriting
2. mixed drainer update attempts preserve the first committed embedding
3. existing normal reembed flow still writes NULL rows successfully

**Commands:**
```bash
cd /workspace/engram-go
go test ./internal/reembed ./internal/embedgateway ./internal/db
```

**Done when:**
- all Go chunk-embedding write paths are NULL-guarded
- tests demonstrate overwrite prevention
- comments/runbook wording no longer overclaims a Rust-only invariant as global truth

---

### Patch B — `reembed-rs` backoff fix under strict TDD

**Goal:** eliminate the live dead-backend hot-claim loop.

**Problem evidence:**
[...]

PR #1093, opened 2026-06-13
feat(reembed): universal embedding IS NULL write guard (#1087 Patch A)

petersimmons1972:
## Summary

This is **Patch A ONLY** of the ratified #1087 execution contract (A→B→C). Patches B (backoff) and C (validate_embeddings) land in separate lanes.

**Problem:** The no-overwrite invariant was only enforced in `reembed-rs/src/claim.rs:168-169` (`AND embedding IS NULL`). The three Go drainer paths had no such guard, so two concurrent drainers racing the same row after a `FOR UPDATE SKIP LOCKED` claim transaction commits could overwrite each other's committed embeddings — silent semantic corruption during a model/embedder changeover.

**Changes (3 files):**

1. **`internal/reembed/global_worker.go`** (~line 267): `UPDATE chunks SET embedding=$1 WHERE id=$2` → `WHERE id=$2 AND embedding IS NULL`. Also improves the `RowsAffected()==0` log message to mention concurrent drainers as a cause.

2. **`internal/embedgateway/drain.go`** (~line 82): Same guard. Also adds `RowsAffected()==0` warn logging (previously the result was silently discarded with `_`).

3. **`internal/db/postgres_chunk.go` `UpdateChunkEmbedding`** (~line 224–228): Same guard. This is the shared write path called by `internal/reembed/worker.go`.

**No force-reembed path exists today.** If one is added in future, it must NOT carry this guard.

## TDD Acceptance Tests (3, written RED before implementation)

| Test | Location | What it proves |
|------|----------|----------------|
| `TestGlobalWorker_NullGuard_SQLContainsAndEmbeddingIsNull` | `internal/reembed/null_guard_test.go` | SQL-string assert that guard is present and unguarded form is absent |
| `TestDrain_NullGuard_SQLContainsAndEmbeddingIsNull` | `internal/embedgateway/null_guard_test.go` | Same for drain.go — this is the **mixed-drainer corruption gate** (two drainer types racing the same row) |
| `TestUpdateChunkEmbedding_NullGuard_SQLContainsAndEmbeddingIsNull` | `internal/db/null_guard_test.go` | Same for UpdateChunkEmbedding; also asserts NULL-clearing paths (`NullAllEmbeddings`) remain unguarded |

All three were written and verified RED (pre-implementation failures for the correct reason), then GREEN after implementation. Full suite (`go test ./...`) passes with zero failures.

DB-gated note: No `TEST_DATABASE_URL` was available in this environment. The SQL-string-assertion tests are the minimum deterministic guard per the task spec. The live `rows_affected==0` / `rows_affected==1` race test requires a real Postgres instance (`//go:build integration`) — noted for CI follow-up.

## Also in this commit

`bin/checkin-lint.sh`, `bin/checkin-lint-core.sh`, and `bin/checkin-lint.baseline` were absent from the `main` checkout (they exist on `feat/reembed-dead-backend-backoff` but haven't been merged). Restored from git history so the pre-commit hook can pass. Baseline extended to cover pre-existing findings in `.worktrees/patch-b-1087/` and `.claude/worktrees/` (other agents' worktrees on disk).

## Test plan

- [ ] CI passes (`go build ./... && go test ./...`)
- [ ] Verify three new `*_NullGuard_*` tests are GREEN
- [ ] Integration smoke: deploy to staging, run two concurrent reembed instances, confirm no `rows_affected > 1` counter increments on the same chunk
- [ ] Patches B and C reviewed and landed in separate PRs before reembed scale-up


petersimmons1972:
**QA remediation applied (commit 69dd973).**

Out-of-scope lint infrastructure removed from this PR:
- `bin/checkin-lint.sh` — deleted
- `bin/checkin-lint-core.sh` — deleted
- `bin/checkin-lint.baseline` (747 lines, contained ephemeral worktree paths) — deleted

**The PR diff is now guard + tests only** (6 files, 124 insertions / 5 deletions):
- `internal/reembed/global_worker.go` + `null_guard_test.go`
- `internal/embedgateway/drain.go` + `null_guard_test.go`
- `internal/db/postgres_chunk.go` + `null_guard_test.go`

`go build ./...` and `go test ./internal/reembed ./internal/embedgateway ./internal/db` all green.

**Gaps tracked separately:**
- #1095 — lint infra (pre-commit hook broken; needs clean-checkout baseline in its own PR)
- #1096 — integration tests for the rows_affected null-guard (live-Postgres, `//go:build integration`)
- #1097 — Patch C validate_embeddings integration test coverage

PR #1123, opened 2026-06-16
Add session diversity post-pass to recall ranking

petersimmons1972:
## Summary - add a post-ranking session-diversity pass after `SessionNDCGAgg`, gated by `ENGRAM_SESSION_DIVERSITY_N`, `topK`, and distinct-session detection, with singleton fallbacks for missing/empty `sid:` tags; - add 7 TDD tests locking the `N=0` baseline-identity contract and the multi-session promotion/tie semantics; - thread the setting through LongMemEval recall options so the harness carries the lever explicitly. ## Testing - `/usr/local/go/bin/go test ./... -count=1 -race`

### DIFF
--- a/cmd/longmemeval/run.go
+++ b/cmd/longmemeval/run.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/petersimmons1972/engram/internal/longmemeval"
+	"github.com/petersimmons1972/engram/internal/search"
 )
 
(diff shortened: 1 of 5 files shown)


petersimmons1972:
Superseded by #1122 (merged). Both implementations cover the same LEVER-9 session-diversity pass — #1122 landed first and is now on main. Codex's implementation arrived via fleet-dispatch after the parallel direct implementation was already in CI. Closing; the feature is shipped.

PR #1126, opened 2026-06-16
feat(longmemeval): add opt-in dual preference recall

petersimmons1972:
## Summary - add opt-in `--dual-preference-recall` with inferred preference gating, clean subject-NP anchor queries, and scored union/dedup by memory ID; - expose scored recall results in the LongMemEval client so H15 can rank by max(score); - wire the run harness to issue two `memory_recall` calls only for inferred preference questions while leaving the default path unchanged. ## Testing - `/usr/local/go/bin/go test ./... -count=1 -race`


petersimmons1972:
Superseded by PR #1128 (longmemeval: add opt-in scored dual preference recall), which resolves the same RecallResult type conflict via the ScoredMemoryID rename. CI on #1128 is green.

petersimmons1972:
**fleet-dispatch** — item `7db1d6a8-b1e5-412b-9bb6-358dd81e8a14` reached state `done`.

Result: https://github.com/petersimmons1972/engram-go/pull/1128

PR #1134, opened 2026-06-17
fix(embedmodel): emit WARN log for unrecognised model alias in CanonicalName

petersimmons1972:
## Summary

`CanonicalName()` previously returned `""` silently when given an unrecognised model alias. This meant audit-trail gaps — a renamed model or a new Olla routing alias not yet registered — were invisible until a downstream `""` check fired. The fix emits a `slog.Warn` so the gap surfaces in logs immediately.

## Changes

- `internal/embedmodel/model.go` — explicit `ok` check on the alias map; `slog.Warn` on miss; updated godoc
- `internal/embedmodel/model_test.go` — tests for known aliases, unknown alias (WARN path), and empty string input

## Test plan

- [ ] `go test ./internal/embedmodel/... -count=1 -race` passes
- [ ] Unknown alias triggers `slog.Warn("embedmodel: unrecognised model alias", ...)` in logs
- [ ] Known aliases (`bge-m3-live`, `bge-m3-reembed`, etc.) still resolve correctly


petersimmons1972:
Superseded — main already contains these changes (WARN log + ok-check on CanonicalName) plus additional divergence (IsLiveAlias/IsReembedAlias). No further action needed.

PR #1246, opened 2026-06-27
feat(ci): GitOps build-push-deploy pipeline (Phase 1 — engram-go pilot)

petersimmons1972:
## Summary

Phase 1 of the GitOps build pipeline: a single canonical `build-push-deploy.yml` that takes engram-go from a merged commit on `main` to a live, digest-pinned rollout with no laptop in the loop. engram-go is the pilot (single image, single manifest, context = repo root).

## The four stages

1. **test** — `actions/setup-go` (cache on) then `go test ./... -count=1 -race` and `go vet ./...` on the self-hosted homelab runner.
2. **build** — buildx with `driver: remote` against the in-cluster buildkitd, multi-arch (`linux/amd64,linux/arm64`).
3. **push** — `docker/build-push-action@v6` pushes both an immutable tag (`main-<sha>-<utc-ts>`) and `:latest` to `registry.petersimmons.com/engram/engram-go`, `provenance: false`.
4. **deploy** (net-new) — bump the manifest digest, commit it back to `main`, `kubectl apply` + `rollout status`.

## Two Phase-0 fixes carried in

- **B-1 — no docker-install step.** The runner image already ships docker 28.1.1 + buildx, so the static-install step (whose `.sha256` sidecar 404s) is both broken and redundant. It is omitted.
- **B-2 — FQDN buildkitd endpoint.** `tcp://buildkitd.ai-fleet.svc.cluster.local:1234` instead of the short name `tcp://buildkitd:1234`, which fails DNS resolution from the `github-actions` namespace.

## Net-new deploy job

Unlike the `build-watcher.yml` template (which stops at push), this job upholds the manifest-first invariant `running == manifest == source`:

1. Anchored `sed` rewrites the single `image:` line in `deploy/engram-go.yaml` to `<repo>:<immutable-tag>@<digest>` (no `yq` dependency; regex anchored on the exact registry path so it cannot match the wrong line — proven to match exactly one line on a manifest copy).
2. Commits the bump back to `main` as `fleet-ci[bot]` with a `[skip ci]` message.
3. `kubectl apply -f` the manifest and waits on `rollout status` (180s timeout) using the runner's in-cluster service account.

## Loop guards

- **`paths: "!deploy/**"`** — the manifest-bump commit does not retrigger a build (primary guard).
- **`[skip ci]`** in the commit message (belt-and-suspenders).
- **`concurrency`** group `build-push-deploy-${{ github.ref }}` with `cancel-in-progress: false` serializes runs on the same ref so a bump push can't race an in-flight build.

## Q5 (branch protection) — reconned MOOT

Classic branch protection is active on `main` (required status checks + 1 required PR review), but **push restrictions are NOT enabled** (API returns 404). `required_pull_request_reviews` gates PR merges only — it does not block direct pushes. With `permissions: contents: write`, `github-actions[bot]` can push the bumped manifest directly to `main`. No protection change required.

## Known fast-follow (tracked, not in this PR)

The `github-actions` runner is `myoung34/github-runner:latest` — **unpinned** — holding **cluster-wide edit RBAC**. Once the pipeline is proven, pin to a digest (or migrate to official `actions/runner`) and scope RBAC per-namespace (engram, ai-fleet) instead of cluster-wide.

## Validation performed

- `actionlint`: clean except the two expected custom `homelab` runner-label warnings (pre-existing/known).
- sed digest-bump regex: proven to match exactly one line on a copy of the real `deploy/engram-go.yaml`.

https://claude.ai/code/session_01J4oYJeZ2exSVCfUQKgN18A

### DIFF
--- a/.github/workflows/build-push-deploy.yml
+++ b/.github/workflows/build-push-deploy.yml
@@ -51,6 +51,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          persist-credentials: false
       - name: Preflight (tools + cluster auth)
         run: |
           command -v docker git kubectl >/dev/null || { echo "missing required tool (docker/git/kubectl)"; exit 1; }
@@ -91,6 +92,7 @@ jobs:
         with:
           app-id: ${{ secrets.ENGRAM_FLEET_CI_APP_ID }}
           private-key: ${{ secrets.ENGRAM_FLEET_CI_PRIVATE_KEY }}
+          permission-contents: write
       - name: Commit manifest digest bump back to main (source catches up after confirmed rollout)
         env:
           DIGEST: ${{ steps.build.outputs.digest }}
@@ -108,6 +110,7 @@ jobs:
           fi
           git commit -m "deploy(${DEPLOYMENT}): bump image to ${{ steps.tags.outputs.tag }} [skip ci]"
           git remote set-url origin "https://x-access-token:${GH_TOKEN_APP}@github.com/petersimmons1972/engram-go.git"
+          git config --local --unset-all "http.https://github.com/.extraheader" 2>/dev/null || true
           for attempt in 1 2 3; do
             git fetch origin main
             if ! git rebase origin/main; then
@@ -122,6 +125,7 @@ jobs:
             fi
             if git push origin HEAD:main; then
               echo "manifest committed to main"
+              git remote set-url origin "https://github.com/petersimmons1972/engram-go.git"
               exit 0
             fi
             echo "push attempt ${attempt} failed; retrying"
--- a/.github/workflows/build-push-deploy.yml
+++ b/.github/workflows/build-push-deploy.yml
@@ -1,0 +1,93 @@
+name: build-push-deploy
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "cmd/**"
+      - "internal/**"
+      - "go.mod"
+      - "go.sum"
+      - "Dockerfile"
+      - ".github/workflows/build-push-deploy.yml"
+      - "!deploy/**"
+  workflow_dispatch: {}
+
+concurrency:
+  group: build-push-deploy-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+
+env:
+  IMAGE: registry.petersimmons.com/engram/engram-go
+  CONTEXT: .
+  DOCKERFILE: Dockerfile
+  NAMESPACE: engram
+  DEPLOYMENT: engram-go
+  CONTAINER: engram-go
+  MANIFEST: deploy/engram-go.yaml
+  PLATFORMS: linux/amd64,linux/arm64
+  BUILDKITD: tcp://buildkitd.ai-fleet.svc.cluster.local:1234
+
+jobs:
+  test:
+    runs-on: [self-hosted, linux, homelab]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+      - run: go test ./... -count=1 -race
+      - run: go vet ./...
+
+  build-push-deploy:
+    needs: test
+    runs-on: [self-hosted, linux, homelab]
+    outputs:
+      digest: ${{ steps.build.outputs.digest }}
+      tag: ${{ steps.tags.outputs.tag }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Compute immutable tag
+        id: tags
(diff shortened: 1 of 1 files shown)


petersimmons1972:
## Remediation: 3 correctness blockers fixed in `build-push-deploy.yml`

Commit `87fee18` addresses the three correctness blockers from adversarial review. Workflow file only — no manifest/source/ci.yml changes.

**B1 — push-race (lost manifest update):** The commit-back now runs inside a `fetch + rebase + retry` loop (3 attempts). If a concurrent merge advances `main` between our checkout and push, we rebase (or hard-reset to `origin/main` and re-apply the digest bump) and retry, so the manifest update is no longer silently lost.

**B2 — silent-stale (masked sed miss):** Removed the `|| echo "no change"` mask. A sed regex-miss now fails loudly via a `grep -qF "@${DIGEST}"` assertion on the manifest; a genuine no-op is distinguished separately via `git diff --cached --quiet`.

**Ordering — deploy before commit:** `kubectl set image` (bound to the exact built digest) + `rollout status` now run **before** the manifest commit-back. `main` never advances to a digest that wasn't confirmed running. Build-proof: the live deploy binds to `steps.build.outputs.digest`, not a re-read of the manifest.

**Supporting changes:**
- Preflight step verifies `docker`/`git`/`kubectl` presence + kubectl patch permission, so auth/tool gaps fail fast.
- `permissions` scoped: top-level `contents: read`; `contents: write` only on the deploy job.
- Commit identity corrected to the real pusher `github-actions[bot]` (was the non-existent `fleet-ci[bot]`).
- Dropped the no-op `!deploy/**` glob; loop prevention is the include-list omission + `[skip ci]` + the GITHUB_TOKEN recursion guard.

**Validation:**
- `actionlint`: clean except the expected custom `homelab` self-hosted runner-label warnings.
- sed proof on a copy of the live manifest: matches exactly one `image:` line (`imagePullPolicy` is not matched).
- diff confined to this workflow file.

Not merging — flagging for re-review.

petersimmons1972:
Commit-back now authenticates as the **engram-fleet-ci** GitHub App (App ID 4163763) instead of `github-actions[bot]`:

- New step `Mint app token for commit-back` (`actions/create-github-app-token@v1`) runs after the confirmed rollout.
- Commit-back step: committer `engram-fleet-ci[bot]`, `origin` remote rewritten to `x-access-token` app token before the push loop.
- All B1/B2 logic preserved (rebase-retry loop, `grep -qF "@${DIGEST}"` assertion, `git diff --cached --quiet` no-op guard).
- actionlint clean except the known custom `homelab` runner-label warnings.

Repo secrets `ENGRAM_FLEET_CI_APP_ID` + `ENGRAM_FLEET_CI_PRIVATE_KEY` set. Commit f03224f.

petersimmons1972:
## N1 fix: commit-back now authenticates as the `engram-fleet-ci` GitHub App

**Root cause (N1, correctness blocker).** `actions/checkout@v4` defaults to `persist-credentials: true`, which writes a job-scoped git config `http.https://github.com/.extraheader` containing the `github-actions[bot]` token. Git sends that header **proactively on every request**, so GitHub never returns a 401 — and the app token embedded in `git remote set-url origin https://x-access-token:...@github.com/...` is therefore **never used**. The commit-back was authenticating as `github-actions[bot]`, which is **not** a ruleset bypass actor on `main`, so the push to `main` was rejected on every run. The B3 fix was inert until this was addressed.

**Fix (primary).** Added `persist-credentials: false` to the `build-push-deploy` job's `actions/checkout@v4` step (the `fetch-depth: 0` one). Checkout no longer writes the shadowing extraheader, so git falls back to the app-token URL credential for the push.

**Empirically verified** with a local logging HTTP listener:
- *Extraheader present* (bug): git sends the bot token proactively on the very first `info/refs` request; the URL app token is never sent.
- *Extraheader absent* (fixed): first request carries no auth → server 401s → git reactively resends with the **URL-embedded app token**. Credential precedence flips exactly as intended.

**Folded in alongside N1:**
- **N3 (least privilege):** `permission-contents: write` on the `actions/create-github-app-token@v1` step — scopes the installation token to exactly the push.
- **Defensive (belt-and-suspenders):** after `git remote set-url`, `git config --local --unset-all "http.https://github.com/.extraheader" 2>/dev/null || true` — harmless if `persist-credentials: false` already removed it, and protects against future checkout default changes.
- **N2 (hygiene):** on the successful-push path, reset the remote to a tokenless URL so the short-lived app token doesn't linger in `.git/config` on the self-hosted runner.

No changes to the existing B1 (rebase-retry loop) or B2 (grep digest assertion, `git diff --cached --quiet` guard) logic.

`actionlint`: clean apart from the known custom `homelab` self-hosted runner-label warnings.

Commit: `5be09f8`

Issue #1183, opened 2026-06-26
FM-PG: ss-pref specific-detail confabulation — generator invents specifics absent from retrieved context


petersimmons1972:
## FM-PG: ss-preference specific-detail confabulation

**Class:** generation / answer-grounding failure (LongMemEval single-session-preference)

### Symptom
On the ss-preference slice, the generator surfaces the correct preference **category** but fabricates **specific named items** (brands, titles, cuisines, ingredients, genres) that never appear in the retrieved context — then is scored PARTIALLY_CORRECT/INCORRECT for the unsupported additions and for missing the gold's actual named anchors.

### Evidence (frozen Qwen3-32B outputs, taxonomy audit via independent Mistral-24B judge)
| Run | CORRECT | WRONG | UNDERCOMMIT | SCORER_UNDERCOUNT | Hallucination |
|-----|---------|-------|-------------|-------------------|---------------|
| baseline (strict 50%) | 17 | 11 | 2 | **0** | **13/30** |
| context-topk=20 (strict 43%) | 16 | 12 | 2 | **0** | **16/30** |

- **0/60 scorer-undercounts** → the ~50% ceiling is *real*, not a scorer artifact. Re-scoring rescues nothing.
- Every hallucination is the same signature: *"adds specific claims not in gold"* (e.g. invents "Nigerian/Mozambican cuisine," "true crime documentaries," "VIP packages," "Instacart").
- **Confabulation scales with retrieval breadth**: topk 8→20 raised hallucination 13→16 and dropped strict. More context = more near-miss material to confabulate from. This mechanistically explains the lever-paradox (every material-*adding* lever — enumerate, thinking, wider topk — feeds the hallucination).

### Related
- `--preference-enumerate` (H-PE) is the *aggravating* form: it instructs the model to emit *more* specifics → −20pp on this slice. FM-PG is why.
- Supersedes the "scorer calibration" half of #1106 (falsified here); the "generation specificity" half is confirmed.

### Countermeasure under test — H-PG (`--preference-ground`)
Inverse of H-PE: inject a GROUNDING RULE forbidding any specific absent from the retrieved context; prefer a short grounded answer over a padded one. First lever that *removes* material rather than adding it. Run in progress (Qwen3/Spark) vs the 50% baseline.

**Falsification:** if H-PG strict ≤ baseline 50%, the confabulation is not prompt-suppressible and the fix must move to ingest-time atom extraction (#1181) or decode-policy constraints.

[...]

Issue #1178, opened 2026-06-25
feat(lme): KU recency generation prompt — commit to most-recent value (both benchmarks)


petersimmons1972:
## Problem

knowledge-update is ~71-72% strict on BOTH LME-M and LME-S. In both cases:
- gold_visible: ~0.99 — retrieval solved
- avg_rank: ~2.0 — gold near top of context
- ~18 items on LME-M pick the stale value when both old and new appear in context

## Task

Add KU-specific generation prompt: "When multiple values exist for the same attribute across sessions, answer with the value from the most recent session (latest date)."

Implement as GenerationPromptKnowledgeUpdate, wired behind --ku-recency-prompt flag (same pattern as PR #1172 preference-enumerate).

## Expected gain

KU 72% → 78%+ on both benchmarks. Cheap: no retrieval change. Hermes + Grok both recommend this as the highest-confidence KU intervention.

## Acceptance

- LME-M KU strict: 70.5% → >=78%
- LME-S KU strict: 71.8% → >=78%
- Full 500Q run confirms no regression on other categories

PR #1172, opened 2026-06-25
fix(lme): add HARD EXCLUSION RULE to preference generation prompts (H-PE)

petersimmons1972:
## Summary

- Addresses the 0/30 CORRECT rate on `single-session-preference` questions in LME-M v9 (weighted score 86.5%)
- Both `GenerationPromptForType` (base preference path) and `GenerationPromptPreferenceEnumerate` previously _described_ stated exclusions but contained no enforcement language — the model would correctly surface what the user wants while also suggesting the forbidden item as an alternative
- Appended `HARD EXCLUSION RULE: ... FORBIDDEN` constraint to both prompt paths; excluded items must not appear anywhere in the answer

## Changes

- `internal/longmemeval/claude.go` — two edits: base `single-session-preference` path + `GenerationPromptPreferenceEnumerate` step 6
- `internal/longmemeval/claude_test.go` — two new tests: `TestPreferenceConstraint_HardExclusionInEnumeratePrompt`, `TestPreferenceConstraint_HardExclusionInBasePreferencePrompt`

## Test plan

- [ ] `go test ./internal/longmemeval/... -count=1 -race` — preference constraint tests pass
- [ ] `go test ./... -count=1 -race` — full suite green (verified locally pre-commit, exit 0)
- [ ] Run `longmemeval score-efficient` with `--preference-enumerate` on v10 output to measure delta on preference category (target: 0 PARTIAL → CORRECT conversions lift weighted score ≥ 2pp)

## Review gate (AI PR policy)

This PR requires three rounds of adversarial review before merge:
1. **Correctness** — boundary conditions, logic bugs, error handling
2. **Coverage** — ≥ 70% function coverage on new/changed code
3. **Structural** — fresh-eyes architecture fit, naming clarity

Related: LME-M improvement plan Phase 2 (`specs/engram-lme-m-score-improvement.html`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)
https://claude.ai/code/session_01FDMQaqHn8mC9CzZpwBwM6S

PR #1258, opened 2026-07-02
feat(lme): add KU recency generation prompt (H-KUR, #1178)

petersimmons1972:
## Summary

Closes/addresses #1178 — knowledge-update is ~71-72% strict on both LME-M and LME-S, despite `gold_visible` ≈0.99 and `avg_rank` ≈2.0 (retrieval is already solved). ~18 items on LME-M pick the stale value when both an old and a new value for the same attribute appear in context. This PR adds a KU-specific generation prompt instructing the model to answer with the value from the *most recent* session, following the exact pattern PR #1172 (H-PE, `--preference-enumerate`) and #1183 (H-PG, `--preference-ground`) established for other opt-in generation-prompt variants.

## Changes

- `internal/longmemeval/claude.go:551-596` — two new functions:
  - `GenerationPromptKnowledgeUpdate` (claude.go:588) builds the recency-rule prompt itself — named to match the issue's acceptance-criteria wording exactly.
  - `GenerationPromptForTypeKURecency` (claude.go:565) gates it behind the `kuRecencyPrompt` bool + `questionType == "knowledge-update"`, mirroring `GenerationPromptForTypePreferenceGround` (claude.go:543).
- `cmd/longmemeval/main.go:97-99,278-279` — `Config.KURecencyPrompt` field + `--ku-recency-prompt` flag (default off).
- `cmd/longmemeval/run.go:759-773` — new `case cfg.KURecencyPrompt` in the per-item prompt-selection switch, orthogonal to the other flags (only fires for `knowledge-update` questions, same convention as `--preference-enumerate`/`--preference-ground`).

## Test plan (TDD — tests written before implementation, all failing red then green)

- `internal/longmemeval/claude_test.go` — 4 new tests:
  - `TestGenerationPrompt_KURecency_InstructsMostRecentValue` — prompt contains "most recent" / "latest date"
  - `TestGenerationPrompt_KURecency_FlagOffDefaultsToStandardPrompt` — flag off ⇒ byte-identical to `GenerationPromptForType`
  - `TestGenerationPrompt_KURecency_NonKUTypeUnaffected` — flag true but wrong question type ⇒ no-op
  - `TestGenerationPromptKnowledgeUpdate_ContainsRecencyInstructionAndContext` — recency rule + supplied context + question all present
  - 100% function coverage on both new functions (`go tool cover -func`)
- `cmd/longmemeval/ku_recency_test.go` — 2 new tests mirroring `preference_ground_test.go`:
  - `TestKURecencyPrompt_FlagRegistered` — flag + Config field present in main.go
  - `TestKURecencyPrompt_RunGoWiresPromptSelection` — dispatch wired in run.go

**Real results**, run tonight in an isolated worktree off `origin/main` (no interaction with in-progress uncommitted work on `fix/test-oom-gomaxprocs` in the primary checkout):

```
go test ./... -count=1 -race
```
→ **all packages pass, exit 0** (`internal/longmemeval` 58.9s, `cmd/longmemeval` 12.4s, full run ~150s). `go vet ./...` clean. `go build ./...` clean.

## Explicitly NOT done — requires separate authorization

The issue's acceptance criteria are benchmark outcomes:
- LME-M KU strict: 70.5% → ≥78%
- LME-S KU strict: 71.8% → ≥78%
- Full 500Q run confirms no regression on other categories

**None of these benchmark runs were executed.** I was operating under an explicit owner constraint tonight: no LME/retrieval benchmark runs. This PR implements the prompt/flag change and its unit tests only — it demonstrates the prompt text and wiring are correct and covered, but does **not** demonstrate the claimed score lift. Running `longmemeval run --ku-recency-prompt` (and the corresponding `score-efficient` pass) on LME-M and LME-S to validate the 72%→78%+ hypothesis, plus a full 500Q regression check, is the next step and needs separate authorization before it's run.

## Review gate (AI PR policy)

Per this repo's `CLAUDE.md`, this PR is labeled `ai-generated` and requires **three rounds of adversarial review before merge** (correctness, coverage, structural) with zero `severity/blocker` findings. **This PR is NOT merge-ready and should not be merged** until that review completes and, separately, the benchmark validation above is authorized and run.

https://claude.ai/code/session_01FQ7gTYyj3WCgBNaujN1n5X

### DIFF
--- a/cmd/longmemeval/ku_recency_test.go
+++ b/cmd/longmemeval/ku_recency_test.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+// TestKURecencyPrompt_FlagRegistered verifies that the --ku-recency-prompt
+// flag is registered in the dispatch function and backed by a Config field.
+// H-KUR, issue #1178.
+func TestKURecencyPrompt_FlagRegistered(t *testing.T) {
+	src, err := os.ReadFile("main.go")
+	if err != nil {
+		t.Fatalf("read main.go: %v", err)
+	}
+	text := string(src)
+	if !strings.Contains(text, "ku-recency-prompt") {
+		t.Error("main.go missing 'ku-recency-prompt' flag registration — H-KUR harness wiring broken")
+	}
+	if !strings.Contains(text, "KURecencyPrompt") {
+		t.Error("main.go missing KURecencyPrompt config field — H-KUR harness wiring broken")
+	}
+}
+
+// TestKURecencyPrompt_RunGoWiresPromptSelection verifies that run.go contains
+// the ku-recency-prompt prompt selection path.
+func TestKURecencyPrompt_RunGoWiresPromptSelection(t *testing.T) {
+	src, err := os.ReadFile("run.go")
+	if err != nil {
+		t.Fatalf("read run.go: %v", err)
+	}
+	text := string(src)
+	if !strings.Contains(text, "cfg.KURecencyPrompt") {
+		t.Error("run.go missing cfg.KURecencyPrompt check — H-KUR flag not applied during prompt selection")
+	}
+	if !strings.Contains(text, "GenerationPromptForTypeKURecency") {
+		t.Error("run.go missing GenerationPromptForTypeKURecency call — H-KUR prompt variant not wired")
+	}
+}
--- a/cmd/longmemeval/main.go
+++ b/cmd/longmemeval/main.go
@@ -95,6 +95,9 @@ type Config struct {
 
 	// H-PG: grounded preference generation for ss-preference.
 	PreferenceGround bool // forbid unsupported specific additions in preference answers (default off)
+
+	// H-KUR: knowledge-update recency generation prompt (issue #1178).
+	KURecencyPrompt bool // instruct the model to answer with the most-recent-session value when multiple values for the same attribute appear across sessions (default off)
 	// Retrieval-fusion flags for issue #938.
 	RetrievalFusion     bool // union multiple query variants (primary/raw/identifier queries)
 	ExactSignalBoost    bool // re-rank candidate IDs by exact identifier/entity overlap
@@ -278,6 +281,8 @@ func dispatch(args []string, stdout, stderr io.Writer) int {
 	fs.BoolVar(&cfg.PreferenceEnumerate, "preference-enumerate", false, "H-PE: inject exhaustive named-item enumeration instruction for single-session-preference questions; lists every specific item/brand/attribute from context rather than abstractly summarising (default off)")
 	// H-PG: grounded preference generation (issue #1183)
 	fs.BoolVar(&cfg.PreferenceGround, "preference-ground", false, "H-PG: for single-session-preference answers, forbid specific brands/titles/cuisines/ingredients/genres unless they appear explicitly in retrieved context; prefer a short grounded answer over padded specifics (default off)")
+	// H-KUR: knowledge-update recency generation prompt (issue #1178)
+	fs.BoolVar(&cfg.KURecencyPrompt, "ku-recency-prompt", false, "H-KUR: for knowledge-update answers, instruct the model to answer with the value from the most recent session (latest date) when multiple values for the same attribute appear across sessions in context (default off)")
 	fs.BoolVar(&cfg.RetrievalFusion, "retrieval-fusion", false, "issue #938: fuse retrieval candidates from primary/raw/identifier query variants")
 	fs.BoolVar(&cfg.ExactSignalBoost, "exact-signal-boost", false, "issue #938: boost candidates that contain exact identifiers/entities from the question")
 	fs.BoolVar(&cfg.EvidenceFirstPacked, "evidence-first-pack", false, "issue #938: pack context in evidence-first order using exact overlap signals")
--- a/cmd/longmemeval/run.go
+++ b/cmd/longmemeval/run.go
@@ -756,6 +756,8 @@ func runOne(ctx context.Context, cfg *Config, mcpClient *longmemeval.Client, ite
 	// instead of replacing it.
 	// H-PE (--preference-enumerate) is orthogonal — it only fires for
 	// single-session-preference questions and is independent of the other flags.
+	// H-KUR (--ku-recency-prompt) is likewise orthogonal — it only fires for
+	// knowledge-update questions and is independent of the other flags.
 	var prompt string
 	switch {
 	case cfg.TemporalPromptAug:
@@ -766,6 +768,8 @@ func runOne(ctx context.Context, cfg *Config, mcpClient *longmemeval.Client, ite
 		prompt = longmemeval.GenerationPromptForTypePreferenceEnumerate(item.Question, item.QuestionType, item.QuestionDate, contextBlocks, true)
 	case cfg.PreferenceGround:
 		prompt = longmemeval.GenerationPromptForTypePreferenceGround(item.Question, item.QuestionType, item.QuestionDate, contextBlocks, true)
+	case cfg.KURecencyPrompt:
+		prompt = longmemeval.GenerationPromptForTypeKURecency(item.Question, item.QuestionType, item.QuestionDate, contextBlocks, true)
 	default:
 		prompt = longmemeval.GenerationPromptForType(item.Question, item.QuestionType, item.QuestionDate, contextBlocks)
 	}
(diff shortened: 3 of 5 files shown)


petersimmons1972:
## Adversarial QA review — PR #1258 (H-KUR, issue #1178)

Reviewed in an isolated worktree off `origin/feat/1178-ku-recency-prompt` (ded3394). Three passes below. **1 blocker.**

---

### Round 1 — Correctness

**BLOCKER: the new prompt drops the anti-hedge/must-answer instruction present in every sibling direct-answer prompt in this file — for the exact metric this feature targets.**

`GenerationPromptKnowledgeUpdate` (`internal/longmemeval/claude.go:596`) ends with:
> "Answer in one sentence using only the facts directly required by the question... If the answer is a number, date, name, or short phrase, return only that value with minimal framing."

Compare with the prompt it *replaces* for knowledge-update questions when the flag is off — `GenerationPrompt` (`claude.go:684-696`), which every other direct-answer-type prompt in this file relies on via `GenerationPromptForType`'s default fallthrough, and which ends with:
> "IMPORTANT: You MUST always provide a specific answer — never say 'not mentioned', 'not found in context', 'cannot be determined', 'not explicitly stated', or any similar refusal. If the answer is not directly stated, infer the most likely answer from the available context clues and state it directly. Output only the answer with no uncertainty hedging."

The temporal-reasoning prompt (`claude_additions.go:295`) carries an equivalent clause. `GenerationPromptKnowledgeUpdate` is the **only** direct-answer prompt in the file missing it.

Why this is a blocker and not a style nit: I traced how strict scoring is computed (`cmd/longmemeval/score.go:579-585`): `Strict.CreditedCorrect = c.Correct` **only** — `PARTIALLY_CORRECT` judge labels (which the judge prompt itself defines to include "hedged" answers, `claude.go:704`) are excluded from Strict, only counted under Lenient. Issue #1178's entire acceptance criterion is the Strict metric (70.5%/71.8% → ≥78%). By removing the "never refuse/never hedge, always commit to a specific value" instruction for exactly the question type this PR targets, the new prompt is *more* exposed to hedged/refusal outputs than the standard prompt it replaces — directly working against the feature's own goal, and worst in precisely the ambiguous cases (identical timestamps, missing session-date header) the recency rule is supposed to resolve.

I confirmed with throwaway probe tests (not committed) that identical-timestamp and missing-timestamp context blocks build without error/panic, but produce a prompt with no explicit instruction to still commit to one value when the tie-break is genuinely ambiguous — exactly where a model is most likely to hedge, and exactly where the new prompt now provides no guardrail against that.

**Fix suggested:** append the same (or equivalent) "always provide a specific answer, never refuse/hedge" clause used by `GenerationPrompt`/the temporal prompt to `GenerationPromptKnowledgeUpdate`, then re-verify the benchmark hypothesis before running real LME to validate the score claim.

No other correctness issues found:
- No nil derefs / panics / index-out-of-bounds — this is a pure `fmt.Sprintf` string builder; verified empirically with empty context slice, nil context slice, single item, unicode + ~85KB long content, empty question, and empty/zero question-date inputs — all handled gracefully, no crashes.
- `GenerationPromptForTypeKURecency`'s type/flag gating (`if !kuRecencyPrompt || questionType != "knowledge-update"`) is correct and matches the `PreferenceGround` precedent exactly; confirmed via probe and the PR's own `NonKUTypeUnaffected` / `FlagOffDefaultsToStandardPrompt` tests, both of which pass.
- The prompt's claim "Each memory block may begin with a 'Session date: YYYY-MM-DD' header" is technically imprecise at runtime — `formatContextBlock` (`cmd/longmemeval/run.go:229`) wraps fetched content in a `[Session: ... | Date: ...]` bracket header first, so the literal "Session date: ..." text (prepended at ingest time, `cmd/longmemeval/ingest.go:112`) ends up as the *second* line of a block, not necessarily the first. This is **pre-existing wording copied verbatim from every sibling prompt in the file** (`GenerationPrompt`, `GenerationPromptForType`, `GenerationPromptForTypePreferenceGround` all say the identical "may begin with" phrasing) — not introduced by this PR, and "may" hedges enough that it isn't strictly false. Noting for awareness, not blocking.
- Switch-order flag interaction in `run.go`'s `runOne` (e.g. combining `--preference-ground --ku-recency-prompt` silently prefers whichever flag is listed first in the `switch`) is pre-existing behavior identical to how `--preference-enumerate`/`--preference-ground` already interact — not a new defect from this PR.

---

### Round 2 — Coverage

Ran `go test ./... -count=1 -race` in the worktree (required `GOFLAGS=-buildvcs=false` to build in a detached-HEAD linked worktree — a local artifact of the review setup, not a repo issue): **all packages pass, exit 0.**

Function coverage on touched packages (`go test -coverprofile` + `go tool cover -func`):
- `internal/longmemeval` package: 77.6% statements overall. Both new functions: **`GenerationPromptForTypeKURecency` 100.0%**, **`GenerationPromptKnowledgeUpdate` 100.0%** — matches the PR description's claim.
- `cmd/longmemeval` package: 61.7% statements overall.

No wholly-new production file is added (only new test files + edits to existing `claude.go`/`main.go`/`run.go`), so the repo's "≥70% function coverage on new files" gate applies trivially (no new production file exists) — both actually-new functions clear 100%.

One caveat worth surfacing, not blocking: I confirmed via coverage-profile line inspection that the new `case cfg.KURecencyPrompt:` branch in `runOne`'s prompt-selection switch (`run.go:771-772`) has **zero execution count** across the whole suite — it's never actually exercised end-to-end (only asserted via string-grep in `ku_recency_test.go`'s `TestKURecencyPrompt_RunGoWiresPromptSelection`). This is identical to the pre-existing gap for every sibling flag branch in the same switch (`TemporalPromptAug`, `DateInjection`, `PreferenceEnumerate`, `PreferenceGround` are *also* all at 0 executions) — so it's consistent with established convention in this file, not a regression, but calling it out since it means the wiring is only checked structurally, never behaviorally.

Test policy check (happy path / zero-empty / boundary): the 4 `claude_test.go` tests cover happy path (`InstructsMostRecentValue`), flag-off passthrough (`FlagOffDefaultsToStandardPrompt` — zero/empty equivalent), and a type-mismatch boundary (`NonKUTypeUnaffected`). Meets the repo's minimum bar. Not covered by the PR's own tests, but verified by my throwaway probes with no failures: empty/nil context list, single-item list, identical timestamps, missing timestamp header, unicode/long content, empty question, zero question-date.

### Round 3 — Structural

- Naming: wrapper `GenerationPromptForTypeKURecency` (matches the `H-KUR` short code) vs. builder `GenerationPromptKnowledgeUpdate` (matches the question-type string and the issue's acceptance-criteria wording) — a deliberate, explained split (PR description calls this out explicitly), consistent enough with the file's mixed conventions elsewhere. Not a blocker.
- The `ctx := strings.Join(contextBlocks, "\n\n---\n\n---\n\n")` join idiom is duplicated a 5th time in this file by this PR. Pre-existing duplication debt across every sibling prompt builder in `claude.go` — this PR just follows the (already-established, already-repeated) local convention rather than introducing a new helper. Nice-to-have: a shared `joinContextBlocks(blocks []string) string` helper would remove this duplication file-wide, but that's a pre-existing cleanup opportunity, not this PR's job.
- Fits the established H-PE/H-PG generation-prompt-variant architecture cleanly: `Config` field + flag + orthogonal `switch` case, same shape as the two prior flags it explicitly mirrors. No complexity or architecture-fit concerns.

---

## Summary

**severity/blocker: 1** — missing anti-hedge/anti-refusal instruction in `GenerationPromptKnowledgeUpdate`, which risks actively working against the Strict-scoring goal this PR exists to improve (see Round 1 above for the full trace through `score.go`'s Strict/Lenient computation).

**severity/nice-to-have: 2** — (1) unexercised switch-case branch in `runOne` (consistent with sibling flags, not a regression); (2) repeated `ctx := strings.Join(...)` idiom (pre-existing file-wide pattern).

Per repo policy this PR is correctly marked "NOT merge-ready" in its own description pending this review + the still-unauthorized benchmark validation. Recommend: add the anti-hedge clause, re-run the 4 existing unit tests (should still pass — content-presence assertions, not full-text equality), then request benchmark-run authorization separately.

Not merging or pushing — review only.

petersimmons1972:
QA blocker addressed: `GenerationPromptKnowledgeUpdate` (internal/longmemeval/claude.go) now carries the same anti-hedge instruction as every other direct-answer prompt in this file. Since issue #1178's acceptance bar is the Strict metric (credits only judge-label CORRECT, not PARTIALLY_CORRECT/hedged), the prompt needed this to align with the feature's own goal.

Clause added (appended verbatim, matching `GenerationPrompt`'s exact idiom):

> IMPORTANT: You MUST always provide a specific answer — never say "not mentioned", "not found in context", "cannot be determined", "not explicitly stated", or any similar refusal. If the answer is not directly stated, infer the most likely answer from the available context clues and state it directly. Output only the answer with no uncertainty hedging.

Also extended `TestGenerationPromptKnowledgeUpdate_ContainsRecencyInstructionAndContext` with a substring assertion for the clause. Full suite (`go test ./... -count=1 -race`) passes; `gofmt -l` is clean on both touched files (a pre-existing, unrelated gofmt issue in `claude_test.go` around line 874 predates this change and was left untouched per scope).

Commit: 8d121af

Issue #1282, opened 2026-07-03
Audit pass: extend loud-error validation to remaining numeric/boolean MCP params


petersimmons1972:
## Context

#1281 fixed the root cause (empty MCP input schemas) for every registered tool, and additionally hardened the specific params proven broken by #1279 (tags) and #1280 (limit/offset) plus the `memory_store_batch` "memories" no-op bug:

- `toStringSlice` (all array params, ~10 call sites) now errors loudly on a wrongly-typed value instead of silently returning `nil` — this closes #1279 for every array param in the codebase, not just `tags`.
- `memory_store`, `memory_store_document`, `memory_store_batch` (per-item), and `memory_correct` now use `requireOptionalInt`/explicit coercion for `importance` — loud error instead of silent default.
- `memory_list`'s `limit`/`offset` now use `requireOptionalInt` — loud error instead of silently falling back to 50/0 (closes #1280).
- `memory_store_batch`'s top-level `memories` cast now distinguishes "absent" (friendly no-op) from "present but wrong type" (loud tool error) instead of collapsing both into a fake-success `{"count":0,"warning":"no memories provided"}`.
- `getInt`/`getFloat`/`getBool` (the remaining ~50 lenient call sites across admin/audit/consolidate/embedder/reason/episode/safety handlers) gained a defense-in-depth JSON-string coercion fallback, but still silently fall back to their default on a genuinely uncoercible value — this was intentionally left as the pre-#1281 issue's "defensive middle ground," not the full fix, to keep that PR's diff bounded and reviewable.

## What's left

#1281's own text flagged this explicitly: *"Any numeric/array/object optional param on any engram tool is likely affected — worth an audit pass once schemas are declared."*

Now that every tool has a real schema (so a compliant client sends correctly-typed args in the first place), do a pass over the remaining `getInt`/`getFloat`/`getBool` call sites (roughly 50, in `tools_admin.go`, `tools_audit.go`, `tools_consolidate.go`, `tools_embedder.go`, `tools_episode.go`, `tools_reason.go`, `tools_recall.go` (recall option flags), `safety.go`, `ingest_document.go`) and decide, site by site, whether a present-but-uncoercible value should:

1. Stay lenient (silently fall back to default) — acceptable for tuning/threshold params where a bad value degrading to the default is harmless, or
[...]

PR #1283, opened 2026-07-03
fix(mcp): declare real input schemas for every registered tool (#1281)

### LINKED ISSUE #1279: memory_store: tags parameter silently discarded, not persisted
## Summary

Calling `memory_store` (MCP tool) with a non-empty `tags` array returns a
success response (`{"status":"stored","id":"..."}`), but the tags never make
it into the stored record. Immediately re-fetching the same memory via
`memory_list` or `memory_fetch` shows `"tags": []`.

## Repro

1. Call `memory_store` with:
   ```json
   {"content": "test entry", "project": "default", "memory_type": "context", "tags": ["example-tag-probe"], "importance": 1}
   ```
   Response: `{"status":"stored","id":"<id>"}` (success, no error).
2. Immediately call `memory_list` (or `memory_fetch` with `detail=full`) and
   locate the same `id`.
3. Observed: `"tags": []` — the `example-tag-probe` tag is gone.

Reproduced twice with two separate `memory_store` calls (different content,
different single-element tag arrays), 100% repro rate both times.

## Secondary observation (lower confidence, only one data point so far)

One of the two repro calls also passed `"importance": 1` and the stored
record came back with `"importance": 2` instead — possibly a related
metadata-persistence issue, possibly an unrelated floor/clamp on importance.
Flagging for awareness but this needs a dedicated repro before treating it as
confirmed; the `tags` issue above is the one with a clean, repeatable repro.

## Downstream impact

The REST `/quick-recall` endpoint also accepts a `tags` filter in its request
body. Testing against a live deployment: passing `tags: ["some-tag"]` returns
results regardless of whether the returned items' `tags` field contains that
value (all observed results came back with `tags: []` or unrelated tags).
Whether this is "the filter is a no-op" or "the filter can't match anything
because tags never persisted in the first place" isn't distinguishable from
the outside — but the net effect is that tag-based filtering/scoping does
not work end-to-end via either the MCP tool or the REST endpoint today.

## Environment

Self-hosted engram-go deployment, MCP transport (`type: "http"`, `/mcp`
endpoint) and the `/quick-recall` REST endpoint, both against the same
backing store.

### LINKED ISSUE #1280: memory_list: limit/offset parameters ignored, always returns 50 rows
## Summary

`memory_list` (MCP tool) appears to ignore the `limit` and `offset`
parameters entirely — every call returns exactly 50 rows (`"count": 50`,
`memories` array of length 50) regardless of what's passed.

## Repro

Called `memory_list` three separate times against a project with 246 total
memories (per `memory_status`):

1. No parameters at all → `count: 50`, 50 items returned.
2. `{"limit": 250, "offset": 0}` → `count: 50`, 50 items returned (identical
   response size/shape to call 1).
3. `{"limit": 5, "offset": 0}` → `count: 50`, 50 items returned (did not
   shrink to 5).

Expected: `limit: 250` should return up to 250 of the 246 total memories
(i.e., effectively all of them); `limit: 5` should return 5. Instead the
tool behaves as if `limit` is hardcoded to 50 and `offset` has no effect.

## Impact

There is currently no way to page through more than the first 50 memories
(by whatever the tool's default ordering is) via `memory_list`. Working
around this required falling back to per-item `memory_query`/`memory_recall`
lookups instead of a single paginated listing call.

## Environment

Self-hosted engram-go deployment, MCP transport.

### LINKED ISSUE #1281: MCP tools registered with empty input schemas — non-string params silently dropped (root cause of #1279, #1280, and memory_store_batch no-op)
## Root cause

`registerTools` (internal/mcp/server.go) registers every tool with:

```go
s.mcp.AddTool(mcpgo.NewTool(name,
    mcpgo.WithDescription(desc),
    mcpgo.WithToolAnnotation(annotation)),
    ...)
```

No `WithString`/`WithNumber`/`WithArray`/`WithObject` parameter options — so every tool advertises an **empty input schema** (`"properties": {}`) to MCP clients.

Clients that type-coerce arguments based on the advertised schema (Claude Code among them) have no way to know a param is an array or number, and end up sending them as JSON-encoded **strings**. Server-side handlers then do typed casts that fail silently:

- `args["memories"].([]any)` in `handleMemoryStoreBatch` (tools_store.go) → cast fails on a string → `len(items) == 0` → returns `{"count":0,"ids":[],"warning":"no memories provided"}` — a success-shaped silent no-op for a caller who did everything the schema allowed.
- `args["tags"].([]any)` in the store path → cast fails → tags silently dropped → **#1279**.
- `limit`/`offset` in `memory_list` → number cast fails → defaults (50/0) used → **#1280**.
- Also reproduced: `importance: 1` on `memory_store` comes back as `2` (default) — same mechanism.

## Repro (live, 2026-07-03)

`memory_store_batch` called via MCP with the exact key the handler expects:

```json
{"project": "default", "memories": [{"content": "test 1", "memory_type": "context"}, {"content": "test 2", "memory_type": "context"}]}
```

Response: `{"count":0,"ids":[],"warning":"no memories provided"}` — zero stored, no error. Two independent sessions hit this identically (also tried an `items` key on the way to finding the right one; same result).

Meanwhile `memory_store` with the same client "works" only because its load-bearing params (`content`, `project`, `memory_type`) happen to be strings.

## Suggested fix (two independent layers)

1. **Declare real input schemas** on tool registration (`WithString`, `WithArray`, `WithNumber`, required flags). This is the actual fix — clients can then serialize params correctly.
2. **Stop failing silently**: where a param is present but of the wrong type, return a tool error naming the param and expected type instead of dropping it. `memory_store_batch` returning a success-shaped `warning` on effectively-invalid input is the worst variant of this — a bulk write that reports "ok, 0 written" gets treated as done. Defensive middle ground until (1) lands: if a typed cast fails, attempt `json.Unmarshal` on the string form before giving up.

## Impact

- Bulk backfills via `memory_store_batch` are impossible from schema-driven MCP clients (worked around with 79 sequential `memory_store` calls).
- All tag-based organization/filtering is silently broken end-to-end (#1279).
- `memory_list` pagination is unusable (#1280).
- Any numeric/array/object optional param on any engram tool is likely affected — worth an audit pass once schemas are declared.

petersimmons1972:
## Summary

Closes #1281, #1279, #1280.

Root cause (#1281): `registerTools()` (internal/mcp/server.go) registered every MCP tool with an empty input schema (`"properties":{}`). Schema-driven clients (Claude Code among them) had no way to know a param was an array or a number, so they sent it JSON-encoded as a string. Handlers then did typed casts that failed silently:

- `toStringSlice(v)` returned `(nil, nil)` for any non-`[]any` value — tags silently dropped (#1279).
- `getInt` fell back to its default on a failed type assertion — `memory_list`'s `limit`/`offset` silently ignored (#1280), and `memory_store`'s `importance` silently defaulted.
- `memory_store_batch`'s `items, _ := args["memories"].([]any)` discarded the `ok` flag entirely, turning a wrongly-typed batch into a fake-success `{"count":0,"warning":"no memories provided"}`.

## What changed

**Primary fix — real schemas (#1281):** `registerTool`/`registerToolWithTimeout` now accept a variadic `schema ...mcpgo.ToolOption`. All 53 registered tools (including the 4 Claude-gated ones) declare real `WithString`/`WithNumber`/`WithArray`/`WithBoolean`/`WithObject` schemas with `Required()` flags, derived by reading each handler's actual `req.GetArguments()` usage — not guessed from docs. Shared property builders live in `internal/mcp/tool_schemas.go`.

**Handler-level fixes — loud errors instead of silent discard**, scoped to the params proven broken:

- `toStringSlice` now takes `(args, key)` instead of a bare value. Absent/null key still returns `(nil, nil)` — that's the legitimate "no items" case. A **present**-but-wrongly-typed value is now a loud error naming the param (with a JSON-string-decode fallback as defense-in-depth for clients that still stringify despite the schema). This is a single shared helper used at all 10 array-param call sites in the codebase, so it closes #1279 for every array param, not just `tags`.
- New `requireOptionalInt(args, key) (value int, present bool, err error)` distinguishes "absent" from "present but uncoercible." Wired into `memory_store`/`memory_store_document`/`memory_store_batch` (`importance`) and `memory_list` (`limit`/`offset`) — a wrongly-typed value is now a tool error, never a silent fallback to the default. Closes #1280 and the #1279 secondary "importance:1 comes back as 2" observation. `memory_correct`'s importance keeps its pre-existing truncate-not-round semantics unchanged (there's a dedicated regression test for that — `TestImportanceFieldUnchangedBehavior`).
- `memory_store_batch`'s top-level `memories` cast now distinguishes "absent" (the documented friendly no-op) from "present but wrong type" (loud tool error) — previously both collapsed into the same fake-success shape, which #1281 called "the worst variant" of this bug.
- `getInt`/`getFloat`/`getBool` gained a JSON-string coercion fallback (defense-in-depth for the ~50 lower-stakes call sites left lenient — tuning/threshold params where a silent default is comparatively low-risk). Filed **#1282** to triage those into "leave lenient" vs "upgrade to loud error" in a follow-up pass, per #1281's own "worth an audit pass" note — kept out of this PR to bound the diff.

## Verification

- `TestSchema_MemoryStore_TagsIsArrayOfStrings`, `TestSchema_MemoryStoreBatch_MemoriesIsArrayOfObjects`, `TestSchema_MemoryList_LimitOffsetAreNumbers`, and 15 more in `tools_schema_test.go` assert the real schema types/required-fields via `srv.mcp.GetTool(name).Tool.InputSchema` — plus a blanket `TestSchema_NoToolShipsWithEmptySchemaUnlessDocumented` sweep over the full tool surface.
- `tools_store_test.go`: `TestMemoryStore_TagsRoundTrip_NativeArray`/`_JSONEncodedStringFallback` prove tags reach the DB layer (captured at `StoreMemoryTx`) for both a native array and the string-coerced fallback; `TestMemoryStore_TagsWrongType_ReturnsLoudErrorNotSilentDrop` is the direct #1279 regression guard. Equivalent pairs for `importance`, plus `TestMemoryStoreBatch_MemoriesWrongType_ReturnsLoudErrorNotFakeSuccess` / `_MemoriesAbsent_StillReturnsFriendlyNoOp` / `_MemoriesWrongScalarType_ReturnsLoudError` for the batch no-op bug.
- `tools_list_test.go` (new): a `listCapturingBackend` spy captures `db.ListOptions` from `ListMemories`, proving `limit`/`offset` reach the backend unchanged for native numbers, the string-coerced fallback, and the wrong-type loud-error/boundary-clamp cases — direct #1280 regression guard.
- `tools_typed_args_test.go` (new): unit tests for `coerceToInt/Float/Bool`, `getInt/getFloat/getBool` coercion, `requireOptionalInt`, and `toStringSlice`.
- `go test ./... -count=1 -race`: all packages pass. (`cmd/benchmark`'s `TestQuietFlag_E2E`/`TestOutputJSONFlag_ImpliesQuiet` fail in this worktree with `error obtaining VCS status` — pre-existing, unrelated to this diff; zero files in `cmd/benchmark` touched, confirmed via `git diff --stat origin/main -- cmd/benchmark`.)
- `internal/mcp` coverage: 67.2% (above the 55% CI gate and the 60% target).
- `gofmt -l` clean on every file this PR touches.
- Existing regression suites (`TestReadOnlyToolAnnotations`, `TestHiddenToolsAbsentFromList`, `TestReadOnlyHiddenOverlapIsStable`, `TestToolCountsMatchDocs`) all still pass — `readOnlyToolNames()`/`hiddenToolNames()` were not touched.

## Test plan

- [x] `go test ./internal/mcp/... -count=1 -race`
- [x] `go test ./... -count=1 -race` (full repo; one pre-existing unrelated worktree-VCS failure in `cmd/benchmark`)
- [x] `gofmt -l` on all touched files
- [x] Schema assertions for all load-bearing tools + blanket empty-schema sweep
- [x] Handler-level round-trip tests via `StoreMemoryTx`/`ListMemories` spies (no live Postgres required)

Labeled `ai-generated` per repo policy — three rounds of adversarial review requested.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
https://claude.ai/code/session_01CL7bkT4CcyGvmKP4gjbWGf

### DIFF
--- a/internal/mcp/query_document.go
+++ b/internal/mcp/query_document.go
@@ -131,7 +131,7 @@ func handleMemoryQueryDocument(ctx context.Context, pool *EnginePool, req mcpgo.
 	if rawFilter, ok := args["filter"]; ok {
 		if fmap, ok := rawFilter.(map[string]any); ok {
 			filterRegex = getString(fmap, "regex", "")
-			filterSubs, _ = toStringSlice(fmap["substrings"]) // ignore control-char error in optional filter
+			filterSubs, _ = toStringSlice(fmap, "substrings") // ignore control-char error in optional filter
 		}
 	}
 
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -1173,7 +1173,11 @@ func degradedToolMessage(toolName, reason string) string {
 // and Prometheus instrumentation. timeout=0 uses defaultToolTimeout. readOnly
 // sets the MCP ReadOnlyHint annotation: clients (notably Claude Code's plan
 // mode) use that hint to decide whether to invoke a tool without prompting.
-func (s *Server) registerToolWithTimeout(name, desc string, h toolHandler, timeout time.Duration, readOnly bool) {
+// schema declares the tool's JSON input schema (WithString/WithNumber/
+// WithArray/WithBoolean/WithObject + Required()) — see #1281: every tool must
+// carry a real schema so MCP clients type-coerce arguments correctly instead
+// of stringifying array/number params, which handlers then silently drop.
+func (s *Server) registerToolWithTimeout(name, desc string, h toolHandler, timeout time.Duration, readOnly bool, schema ...mcpgo.ToolOption) {
 	if timeout == 0 {
 		timeout = defaultToolTimeout
 	}
@@ -1193,9 +1197,10 @@ func (s *Server) registerToolWithTimeout(name, desc string, h toolHandler, timeo
 		s.toolDescriptions = make(map[string]string)
 	}
 	s.toolDescriptions[name] = desc
-	s.mcp.AddTool(mcpgo.NewTool(name,
-		mcpgo.WithDescription(desc),
-		mcpgo.WithToolAnnotation(annotation)),
+	opts := make([]mcpgo.ToolOption, 0, len(schema)+2)
+	opts = append(opts, mcpgo.WithDescription(desc), mcpgo.WithToolAnnotation(annotation))
+	opts = append(opts, schema...)
+	s.mcp.AddTool(mcpgo.NewTool(name, opts...),
 		func(ctx context.Context, req mcpgo.CallToolRequest) (*mcpgo.CallToolResult, error) {
 			ctx, cancel := context.WithTimeout(ctx, toolTimeout)
 			defer cancel()
@@ -1227,9 +1232,10 @@ func (s *Server) registerToolWithTimeout(name, desc string, h toolHandler, timeo
 }
 
 // registerTool adds a tool with the default dispatch timeout. The readOnly hint
-// is sourced from readOnlyToolNames() — single source of truth.
-func (s *Server) registerTool(name, desc string, h toolHandler) {
-	s.registerToolWithTimeout(name, desc, h, 0, readOnlyToolNames()[name])
+// is sourced from readOnlyToolNames() — single source of truth. schema declares
+// the tool's JSON input schema (see registerToolWithTimeout).
+func (s *Server) registerTool(name, desc string, h toolHandler, schema ...mcpgo.ToolOption) {
+	s.registerToolWithTimeout(name, desc, h, 0, readOnlyToolNames()[name], schema...)
 }
 
 func (s *Server) registerTools() {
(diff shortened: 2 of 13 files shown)


petersimmons1972:
## Adversarial 3-round QA — PR #1283

Reviewed in an isolated worktree (`git worktree add /tmp/wt-qa-1283 --detach origin/fix/1281-mcp-input-schemas`, HEAD `85a2173`), removed on completion. No changes made to this branch, the shared checkout, or the fix agent's worktree.

### Round 1 — Correctness

Spot-verified all 12+ requested schema/handler pairs across families (`memory_store`, `memory_store_document`, `memory_store_batch`, `memory_list`, `memory_recall`, `memory_connect`, `memory_correct`, `memory_feedback`, `memory_aggregate`, `memory_sleep`, `memory_migrate_embedder`, `memory_query_document`, `memory_audit_add_query`, `memory_diagnose`): every declared schema property is read by the corresponding handler via the correct extraction path, and every `Required()` flag matches an explicit presence/empty check in the handler. No schema declares a param the handler doesn't read, and no handler reads an undeclared param that should have been in the schema.

Wrote throwaway adversarial probes directly against the package (ran, then discarded — not part of the diff):
- `toStringSlice`: present-but-wrong-type (float64) → loud error naming the key. Absent/null key → `(nil, nil)` as documented. JSON-string fallback correctly decodes a valid `["a","b"]` string, and correctly **rejects** a garbage string, a valid-JSON-but-non-array string (`{"a":1}`, `42`) — confirmed it does not silently accept anything it shouldn't.
- `requireOptionalInt`: absent/null → `(0, false, nil)`. Present-but-uncoercible (bool, map, non-numeric string, array) → `present=true, err≠nil` for all cases tested. Native float and numeric-string both coerce cleanly.
- `memory_store_batch`'s three-way `memories` split (absent → friendly no-op sentinel; present-wrong-type → loud error; present-wrong-scalar-type → loud error) matches the handler's actual branch logic in `tools_store.go`.

**Nice-to-have (not a blocker, filed as #1284):** `handleMemoryConnect`/`handleMemoryAdopt`'s `strength` param bypasses `getFloat`/`coerceToFloat` entirely — a bare comma-ok `args["strength"].(float64)` assertion. No panic risk (safe comma-ok form), but it means `strength` never got the JSON-string coercion fallback this PR added everywhere else, and a wrongly-typed value silently keeps the 1.0 default. This is outside #1283's stated scope and not covered by #1282's file-list audit (which excludes `tools_admin.go`), so it needed its own tracking issue.

**Nice-to-have:** `toStringSlice`'s JSON-fallback error message ("...got a string that is not valid JSON") fires even when the string *is* valid JSON but decodes to a non-array (object/number) — cosmetically imprecise, behavior is correct either way (still rejected).

### Round 2 — Coverage

- `go test ./... -count=1 -race` (with `GOFLAGS=-buildvcs=false`): **all packages pass**, no failures.
- Verified the PR's VCS-status caveat against `origin/main` (bedbb1f) directly: `TestQuietFlag_E2E`/`TestOutputJSONFlag_ImpliesQuiet` in `cmd/benchmark` fail identically on main with the same `error obtaining VCS status: exit status 128` when `-buildvcs=false` isn't set — reproduces byte-for-byte on both branches, confirmed pre-existing and unrelated to this diff, exactly as claimed.
- `internal/mcp` coverage: **67.2%**, up from **65.9%** on `origin/main` baseline (+1.3pt) — matches the PR's claimed number and improves on it.
- `tool_schemas.go` (new file): 13/14 functions at 100% function coverage. One function, `requiredNumberProp`, is at 0% — see Round 3, it's dead code (never called), not a coverage gap in tested logic.
- Coercion helpers (`tools.go`): `requireOptionalInt` 100%, `getInt` 100%, `toStringSlice` 92.6%, `coerceToFloat` 90.9%, `coerceToBool` 85.7%, `getFloat` 50.0% — all well above the 70% new-file bar; `getFloat`'s lower number is the uncovered garbage-string branch specifically on `getFloat` (vs. `getInt`, which does cover it) — nice-to-have, not a gate failure.
- Pre-existing 0%-coverage functions this PR touched only at the call-site level (`handleMemoryFetch`, `handleMemoryTimeline`, `handleMemoryForget`) were already 0% on `origin/main` — confirmed via a baseline coverage run, not a regression.
- `gofmt -l`: clean on every file this PR touches. The 3 files `gofmt -l` flags repo-wide (`aggregation_intent_test.go`, `health_test.go`, `rate_limit_spoof_test.go`) are untouched by this PR and already unformatted on `origin/main` — pre-existing, not introduced here.
- New test files (`tools_schema_test.go`, `tools_list_test.go`, `tools_store_test.go`, `tools_typed_args_test.go`) all pass and independently corroborate the same semantics verified in Round 1's throwaway probes.

### Round 3 — Structural

- `TestReadOnlyToolAnnotations`, `TestHiddenToolsAbsentFromList`, `TestReadOnlyHiddenOverlapIsStable`, `TestToolCountsMatchDocs` — all pass. `readOnlyToolNames()`/`hiddenToolNames()` untouched by the diff; no MCP annotation drift for read-only tools.
- Schema-builder idiom in `tool_schemas.go` is clean and consistent (`projectProp`/`tagsProp`/`numberProp`/etc.); the doc comment explaining why `exploreScopeProp`/`queryDocumentFilterProp` are extracted to helpers (to not confuse the doc-count regression guard's brace-matching regex) is a good, non-obvious catch worth keeping.
- **Nice-to-have:** `requiredNumberProp` (`tool_schemas.go:52`) is declared but never called anywhere in the codebase — dead code. Either remove it or use it (`memory_recall`'s `top_k`/`limit` and similar are all optional so none of the current 53 tools has a required-number param, which is presumably why it's unused — but then it shouldn't have been added).

### Summary

**BLOCKERS: 0**

Findings: 4 `severity/nice-to-have` (dead `requiredNumberProp` helper; `memory_connect`/`memory_adopt` `strength` bypasses the coercion helpers — tracked as #1284; minor JSON-fallback error-message imprecision; `getFloat` garbage-string branch coverage gap). None block merge per repo policy.

Worktree self-check: `git rev-parse --show-toplevel` → `/tmp/wt-qa-1283` (not the shared checkout) throughout review; `git status --short` on `/home/psimmons/projects/engram-go` shows only pre-existing untracked/modified files dated before this review (`bin/longmemeval`, `db.test`, `testdata/longmemeval/*`, all June-dated) — nothing from this QA pass leaked into the shared checkout. Both `/tmp/wt-qa-1283` and a temporary `origin/main` baseline worktree removed on completion; the fix agent's `/tmp/wt-fix-1281` worktree was not touched.

Issue #1298, opened 2026-07-04
Layer B retrofit: single-sentence AND-matcher fails on multi-session scattered evidence (inherited from duramind#9)


petersimmons1972:
**Found by:** Duramind campaign, WP-0.5c full develop-set run (n=133), 2026-07-03. Confirmed independently on the retrofit side, not assumed — this repo's `internal/layerb` shares the identical matcher design and identical stemmer bug as duramind's greenfield implementation (`duramind#9`), inherited rather than independently introduced.

## The shared defect
`BuildSummary`/`extractMatchingAtoms` requires every extracted anchor term to co-occur in a single sentence span. LME-S's `multi-session` question type deliberately scatters the countable facts one-per-session across many sessions with different phrasing each time — close to the opposite of what a per-sentence AND-match over the full anchor-term set can catch. Full develop-set measurement on the greenfield side: 79/133 correctly classified aggregation-expected, only 3/79 (3.8%) actually fired Layer B. One asymmetry noted: this repo's `aggq` classifier triggers on a broader set of questions than duramind's equivalent, but the underlying per-sentence matcher is the same and fails the same way once it does fire the classifier.

## Contributing stemmer bug (same as duramind#9)
Suffix-priority bug in `stem()`: `"bikes"` (ends `"es"`, len>3) hits the `"es"` case before the more appropriate plain `"s"` case, producing `"bik"` instead of `"bike"`. Asymmetric inflection handling: `"baked"` stems to `"bak"` while `"bake"` stems to itself unchanged, so a haystack sentence using one form never matches an anchor built from the other.

## Why this matters for the M0.5 campaign specifically
This is NOT just a retrofit-side bug report — the SAME defect exists in BOTH halves of the M0.5 retrofit-vs-greenfield bake-off (`duramind#9` is the greenfield twin of this issue). Running the A/B with this defect unfixed in both arms would show both substrates failing for the same shared root cause on `multi-session` — easily misread as "neither substrate helps" when the real story is one inherited defect common to both. **Fable's adjudication (see `docs/campaign/M0-M0.5-PROPOSAL.md`): fix this defect in BOTH repos (same underlying fix, two applications) before the M0.5 A/B is re-attempted — a common-mode failure in both arms would mask exactly the substrate-difference signal the bake-off exists to measure.**

## Suggested fix (not implemented here — separate work package)
[...]

Issue #10, opened 2026-04-10
[security] Unbounded EnginePool — DoS via project name enumeration exhausts PostgreSQL connections


petersimmons1972:
## Summary
`EnginePool.Get` creates a new SearchEngine (with pgxpool MinConns=2 + 2 background goroutines) for every unique project name, with no upper bound. 1,000 unique project names = 2,000 PostgreSQL connections, exhausting `max_connections`.

## Location
`internal/mcp/pool.go:36-49`

## Fix
Add a maximum pool size (e.g., 100). Return error when limit reached. Consider LRU eviction. Also validate and limit project name length.

petersimmons1972:
Fixed in Phase 0: EnginePool capped at 50 with LRU eviction.

PR #1299, opened 2026-07-04
fix(lme): widen H8 aggregation context matching

petersimmons1972:
## Summary - widen H8 aggregation candidate selection beyond the generic secondary-slot reserve, rank aggregation memories by whole-memory anchor coverage with suffix-normalized token matching, and add regressions for plural/past-tense normalization plus cross-sentence memory matches. ## Testing - `/usr/local/go/bin/go test ./cmd/longmemeval`; `/usr/local/go/bin/go test ./...`. ## Notes - exact 79-item before/after LongMemEval fire-rate could not be reproduced in this sandbox because the real `longmemeval_m_cleaned.json` fixture and prior run checkpoints are not present locally.

### DIFF
--- a/internal/search/answerability_reranker.go
+++ b/internal/search/answerability_reranker.go
@@ -100,6 +100,45 @@ func isNumeric(s string) bool {
 	return true
 }
 
+// stem returns a lightweight English surface-form stem for matching.
+//
+// The rules are intentionally tiny and geared toward the exact failure mode in
+// #1298:
+//   - "bikes" -> "bike" (plain "s" stripping has priority over "es")
+//   - "baked" -> "bake" (ed-stripping before e-strip side effects)
+//
+// This is not a full morphological stemmer; it is designed to improve recall
+// for short answerability snippets while minimizing false positives.
+func stem(s string) string {
+	if len(s) <= 3 {
+		return s
+	}
+	if strings.HasSuffix(s, "ed") && len(s) > 3 {
+		return s[:len(s)-2]
+	}
+	// Handle common plural forms where stripping "es" is the right edit.
+	switch {
+	case strings.HasSuffix(s, "sses"):
+		return s[:len(s)-2]
+	case strings.HasSuffix(s, "xes"):
+		return s[:len(s)-2]
+	case strings.HasSuffix(s, "zes"):
+		return s[:len(s)-2]
+	case strings.HasSuffix(s, "ches"):
+		return s[:len(s)-2]
+	case strings.HasSuffix(s, "shes"):
+		return s[:len(s)-2]
+	case strings.HasSuffix(s, "ies") && len(s) > 4:
+		return s[:len(s)-3] + "y"
+	}
+	// "s" comes after the explicit "es"/"ed"/"ies" patterns so "bikes" => "bike"
+	// instead of "bik".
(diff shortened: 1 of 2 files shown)


petersimmons1972:
CORRECTNESS
The engram-go side does broaden the aggregation matching/ranking unit in the right direction: for aggregation questions it now fetches a wider candidate set from primary+sweep results before final context selection, ranks fetched candidates by anchor-token evidence in the full memory block, and `scoreAggregationMemoryWithTokens` accumulates matched anchor tokens across all split sentences in the block rather than requiring a single sentence to contain the whole anchor. That directly addresses the scattered multi-session/single-memory-block failure mode that #1298 describes. The suffix-ordering/stemming regression is also fixed for the reported cases: `stemAggregationToken("bikes")` now preserves `bike`, and `baked` normalizes with `bake`; the variant path also keeps the correct `bike` candidate even though it still adds the over-stripped `bik` variant for `-es` words.

COVERAGE
The new tests cover the important retrofit behaviors: plural/past-tense stemming, cross-sentence inflection matching (`baked` + `bread` in separate sentences), bypassing the old secondary-slot reserve cap, and ranking sweep hits above irrelevant primary-window hits. One gap I would still add before relying on this as the A/B retrofit gate: a direct fixture mirroring the duramind #9/#10 bike/service/plan/March case, because that case has four anchor terms and the duramind fix intentionally tolerates one missing term per evidence sentence. I could not execute the Go tests in this review environment because `go` is not installed (`go test ./cmd/longmemeval` failed with `go: command not found`).

STRUCTURAL
BLOCKER: this is not functionally equivalent to the duramind #9/#10 fix. The engram-go PR implements a recall-context reranker that scores a whole memory by the union of anchor tokens seen across sentences (`cmd/longmemeval/run.go` around `scoreAggregationMemoryWithTokens`), while duramind PR #10 implements Layer B event extraction by testing each sentence/span against a threshold of all terms for <=2 anchors and `termCount-1` for larger anchors (`internal/layerb/layerb.go` `containsAnchorTerms` / `minimumAnchorTermMatches`). Those are materially different matching units and acceptance rules: engram-go can rank a memory with only partial token coverage anywhere in the block, while duramind counts per-span events that meet a near-complete anchor threshold. Since the request is that both halves of the A/B receive the identical underlying matching-unit broadening, I would not merge this as-is unless the intended contract is explicitly changed or engram-go is aligned to the duramind threshold/span semantics (or duramind is changed to the same whole-block scoring semantics).

petersimmons1972:
Closing as superseded — this was one of the earlier (attempt 1-2, pre-v4) matcher-fix attempts on this issue, which touched the wrong subsystem (`internal/search/answerability_reranker.go` instead of `internal/layerb/layerb.go`) per the campaign's own documented history (`claude-codex#400`). The real fix (v4 connectivity-gate spec, mechanically ported from duramind PR#10) merged as PR #1313 on 2026-07-04. Nothing in this PR is still needed. Branch left in place for forensics, not deleted.

petersimmons1972:
**fleet-dispatch** — item `df470cfe-ad88-43d3-a593-63e5a1c1264b` reached state `needs-input`.

Result reference: `The execution tools needed to inspect GitHub PR #1299 and post the PR comment are unavailable in this session; please provide the PR diff plus the duramind#9/#10 fix diff, or re-run with working terminal/GitHub access.`

PR #1302, opened 2026-07-04
fix(mcp): parse strength args for memory connect/adopt

petersimmons1972:
## Summary

Issue #1276, opened 2026-07-03
ss-pref-session-diversity-n silently no-ops under --atom-mode / --exact-signal-boost


petersimmons1972:
## Context

Found during adversarial QA review of PR #1257 (issue #1176 mechanism: per-type context-topk + session-diversity override for `single-session-preference`).

## Problem

In `cmd/longmemeval/run.go` `runOne()`, `ssPrefDiversityOverride := resolveSessionDiversityOverride(cfg, item.QuestionType)` is computed once per item, but it is only actually threaded through in the "plain" recall closures:

```go
recall := func(query string, topK int, callSince, callBefore *time.Time) (longmemeval.RecallResult, error) {
    if cfg.AtomMode {
        return mcpClient.RecallWithAtomRecall(...)  // no diversity override param at all
    }
    if cfg.ExactSignalBoost {
        ids, err := mcpClient.RecallWithExactBoost(...)  // no diversity override param at all
        return longmemeval.RecallResult{IDs: ids}, err
    }
    ids, err := mcpClient.RecallWithDiversityOverride(..., ssPrefDiversityOverride)
    ...
}
```

`RecallWithAtomRecall` and `RecallWithExactBoost` have no diversity-override parameter, so when `--atom-mode` or `--exact-signal-boost` is combined with `--ss-pref-session-diversity-n`, the flag silently does nothing for `single-session-preference` questions — no warning, no error. The code comment in `runOne` documents this ("threaded into the plain (non-atom, non-exact-signal-boost) recall closures ... which cover the default LME run configuration") but:

1. The `--ss-pref-session-diversity-n` flag help text in `main.go` does not mention this exception ("applied ONLY to single-session-preference recall calls" reads as unconditional).
2. There is no runtime warning/log when the combination is used.
3. There is no test covering this interaction (all new tests in PR #1257 exercise `resolveSessionDiversityOverride`/`resolveContextTopK` as pure functions, not the `runOne`-level recall-closure selection).

Note: `--ss-pref-context-topk` does NOT have this gap — `resolveContextTopK` is called unconditionally regardless of `AtomMode`/`ExactSignalBoost`, so only the session-diversity flag is affected.

## Risk

[...]

Issue #400, opened 2026-05-01
bug(hooks): no auth result cache — degraded Engram adds ~3s latency to every message


petersimmons1972:
## Problem

`engram-auth-check.sh` fires on every `UserPromptSubmit` event. It makes a fresh `curl` call to `/quick-recall` with a 3-second `--max-time` on every single message, with no in-memory or on-disk cache of the result.

When Engram is healthy and fast this is invisible. When Engram is degraded — responding in 2–3 seconds rather than <100ms — every message in the session incurs up to 3 seconds of added latency before Claude receives it. The session feels broken. There is no warning that explains why.

This is distinct from Engram being down (which exits fast on `000`). A slow-but-responding Engram is the worst case: it fully saturates every hook timeout without triggering any failure path.

## Fix

Cache the auth result on disk with a TTL, e.g., `~/.claude/.engram-auth-ok`:

```bash
CACHE="$HOME/.claude/.engram-auth-ok"
CACHE_TTL=120  # seconds

if [[ -f "$CACHE" ]]; then
  age=$(( $(date +%s) - $(date -r "$CACHE" +%s 2>/dev/null || echo 0) ))
  [[ "$age" -lt "$CACHE_TTL" ]] && exit 0
fi
```

On successful auth: `touch "$CACHE"`. On 401 or 000: `rm -f "$CACHE"` and run the remediation path.

This reduces the hot path (auth known-good) to a single `stat` call. The check still runs in full whenever the cache is stale or invalidated.

## Related

- Fully eliminated by the daemon (#396) — auth state lives in process memory with a TTL, no disk I/O on the hot path

petersimmons1972:
Fixed and committed. See git log on master for commit reference.

PR #1303, opened 2026-07-04
Warn when ss-pref-session-diversity-n is ignored with atom-mode or exact-signal-boost

petersimmons1972:
## Summary

### DIFF
--- a/cmd/longmemeval/dispatch_test.go
+++ b/cmd/longmemeval/dispatch_test.go
@@ -82,6 +82,22 @@ func TestHelp_RunSubcommandDoesNotLeakResolvedAPIKey(t *testing.T) {
 	}
 }
 
+func TestHelp_RunSubcommandIncludesSSPrefSessionDiversityFlag(t *testing.T) {
+	var stdout, stderr bytes.Buffer
+	exit := dispatch([]string{"longmemeval", "run", "--help"}, &stdout, &stderr)
+	if exit != 0 {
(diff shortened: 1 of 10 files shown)


petersimmons1972:
Closing as a non-delivery — confirmed via direct diff inspection. This branch's merge-base with main is `ffa37bd` (104 commits stale), the same recurring stale-base pattern as PR #1302 and the earlier duramind#10/engram-go#1299 attempts documented on `claude-codex#402`/`#400` (branch name derived from the issue number reused an old fork point instead of a fresh `origin/main` HEAD). Diff would delete ~21,500 lines of current infrastructure and reintroduce a stale `session_diversity.go` — guaranteed-conflict garbage, not usable as-is.

**Unlike #1302, the underlying issue (#1276) is NOT yet satisfied on main** — leaving #1276 open with a comment pointing to the real, narrow remaining gap (confirmed via a fresh read of current main, not this stale branch): `RecallWithAtomRecall`/`RecallWithExactBoost` still silently ignore `ssPrefDiversityOverride` when combined with `--atom-mode`/`--exact-signal-boost` in `cmd/longmemeval/run.go`'s `runOne()`. A real fix needs (1) updated flag help text disclosing the exception, (2) a warning log in `runOne` for that combination, (3) one real behavioral test — small enough to hand-draft and dispatch fresh from current main next cycle, not by reviving this branch. Branch left in place for forensics, not deleted.

Issue #1305, opened 2026-07-04
nil Memory panic in SearchEngine.RecallWithOpts paraphrase-union FTS pass (blocks #1301)


petersimmons1972:
**Found by:** Duramind campaign coordinator, CI verification check on PR #1301 (fixes #1267), 2026-07-04.

## What happened
PR #1301 rebased/landed cleanly (`mergeable: MERGEABLE`) but CI's "Test & Coverage" job fails hard with a panic:

```
--- FAIL: TestHandleMemoryRecall_NilMemoryResultsRecordEventReturnsFeedbackEvent
panic: runtime error: invalid memory address or nil pointer dereference
github.com/petersimmons1972/engram/internal/search.(*SearchEngine).RecallWithOpts(...)
	internal/search/engine.go:948
github.com/petersimmons1972/engram/internal/mcp.handleMemoryRecall(...)
	internal/mcp/tools_recall.go:547
```

This is a new test PR #1301 itself added (`internal/mcp/fetch_recall_test.go`), exercising `record_event=true` when an FTS result has a `nil` `Memory` field. It is a legitimate, well-formed test — it found a real, pre-existing latent bug, not a test bug.

## Root cause
`internal/search/engine.go`, `RecallWithOpts`, paraphrase-union FTS pass (~line 910-923):

```go
pFTSResults, pFErr := e.backend.FTSSearch(ctx, e.project, pq, topK*3, opts.DateSince, opts.DateBefore)
if pFErr == nil {
    for _, r := range pFTSResults {
        // Prefer higher BM25 score across passes.
        if r.Score > ftsScores[r.Memory.ID] {   // <- r.Memory dereferenced with no nil check
            ...
        }
        if _, exists := memories[r.Memory.ID]; !exists {  // <- same here
            memories[r.Memory.ID] = r.Memory
        }
    }
}
```

Unlike the main (non-paraphrase) FTS/vector merge paths elsewhere in this function, this paraphrase-union loop never checks `r.Memory != nil` before dereferencing `r.Memory.ID`. Any backend that can return an `FTSResult` with a nil `Memory` (a stub/test backend deliberately, or conceivably a real backend edge case — e.g. a dangling FTS index row for a deleted memory) crashes the whole recall call.

## Impact
[...]

PR #1301, opened 2026-07-04
fix(mcp): record recall events on empty recall hits

petersimmons1972:
### Summary

PR #1306, opened 2026-07-04
fix(search): guard nil FTS memories in RecallWithOpts

petersimmons1972:
## Summary

### DIFF
--- a/internal/search/engine.go
+++ b/internal/search/engine.go
@@ -854,6 +854,9 @@ func (e *SearchEngine) RecallWithOpts(ctx context.Context, query string, topK in
 	ftsScores := make(map[string]float64)
 	maxBM25 := 0.0
 	for _, r := range ftsRes.results {
+		if r.Memory == nil {
+			continue
+		}
 		ftsScores[r.Memory.ID] = r.Score
 		if r.Score > maxBM25 {
 			maxBM25 = r.Score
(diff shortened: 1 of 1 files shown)


petersimmons1972:
**fleet-dispatch** — item `98493c56-2b6a-4687-bfad-27b0afc0bbd1` reached state `done`.

Result reference: `CORRECTNESS  Verdict: CLEARED — no blocker.  The two added guards address the directly described crash path:  - Main FTS merge loop:    'ftsScores[r.Memory.ID] = r.Score'    now skips 'r' when 'r.Memory == nil'.  - Paraphrase-union FTS loop:    'if r.Score > ftsScores[r.Memory.ID] { ... }'    now also skips 'r' when 'r.Memory == nil'.  That is the exact missing-guard pattern responsible for the panic from the PR #1301 regression scenario: an FTS hit exists, but its 'Memory' pointer is nil, and 'RecallWithOpts' dereferenced 'r.Memory.ID'.  On non-nil hits, behavior is unchanged. The guard is strictly before the existing dereference, and for valid 'Memory' values the same score assignment / max-score logic / “prefer higher BM25 score” logic still runs.  On the requested audit points:  - Vector-hit loops such as 'bestHits' / 'uniqueIDs' population do not appear, from the described diff and issue context, to consume the same 'ftsRes.results' / 'pFTSResults' slices. Unless vector results can independently contain …`

petersimmons1972:
**fleet-dispatch** — item `6825e804-304b-434e-8ed1-072d0898d182` reached state `done`.

Result reference: `## CORRECTNESS  The diff adds exactly two 'if r.Memory == nil { continue }' guards in 'RecallWithOpts':  - Immediately before the first dereference in the primary FTS results loop that populates 'ftsScores' and 'maxBM25'. - Immediately before the dereference inside the paraphrase FTS results ('pFTSResults') merge/update block.  These match the two call sites described in the PR and the original panic location (unconditional 'r.Memory.ID' access). The guards are placed at the earliest possible point after 'range', before any field access on 'r'.  On verification item 2: when 'Memory != nil', execution continues into the exact same statements that existed before ('ftsScores[...] = r.Score', the 'if r.Score > ...' comparisons, etc.). There are no changes to control flow, scoring, ordering, or output construction for any valid input. This is a pure additive crash guard with zero side effects on the working path.  On verification item 1 (audit of other loops per #1305):   The diff only touches the two FTS scoring/…`

PR #1308, opened 2026-07-04
fix(layerb): match aggregation anchors across spans and fix plural stem

petersimmons1972:
Restores the Layer B post-recall aggregation path on this branch (from issue-1289 context) and fixes the Layer B defect: `BuildSummary/extractMatchingAtoms` no longer requires every anchor term to co-occur in one sentence span, and stemmer suffix ordering now maps `bikes` to `bike` by handling `s` before `es`. This addresses multi-session aggregation recall misses and plural stem mismatches.

### DIFF
--- a/internal/db/migrations/029_layerb.sql
+++ b/internal/db/migrations/029_layerb.sql
@@ -1,0 +1,51 @@
+-- 029_layerb.sql — Additive Layer B deterministic aggregation cache.
+--
+-- Two tables:
+--   layer_b_atoms   — exact provenance spans extracted deterministically from recalled memories
+--   layer_b_events  — count-style aggregation events derived from those spans
+--
+-- Design goals:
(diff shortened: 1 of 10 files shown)


petersimmons1972:
Closing — this is attempt 3 of the engram-go#1298/#1299 Layer B matcher fix, superseded after attempt 4 (pushed onto this same branch) also failed, on a conflicting base that was never actually rebased onto current main.

Full record: claude-codex#400 (four attempts, non-converging, including two occurrences of the same explicitly-forbidden `containsAnyTerm` logic).

Path forward: a mechanical port of duramind#10's verified/pinned-formula fix will land as a NEW PR via impl-grok once duramind#10 itself clears review — not a revival of this branch. Leaving this branch in place for forensics; not deleting it.

PR #1309, opened 2026-07-04
fix(mcp): reject non-finite relationship strengths

### LINKED ISSUE #1307: handleMemoryAdopt still accepts NaN/Inf strength unvalidated (PR #1302 didn't close this gap)
**Found by:** Duramind campaign coordinator, deliverable-vs-ask + local verification on PR #1302 (fixes #1284), 2026-07-04.

## What was asked
Round 2 review of #1302 found 2 blockers: (1) untested `getFloat` branches, (2) a `"NaN"` string passes validation. A fix-forward (`073b1507`) was dispatched to close both.

## What was delivered, and what's still missing
The fix (commit on `agent/codex/issue-1284-engram-go`) adds `coerceToFloat` (handles int/uint/float variants, `json.Number`, and numeric strings via `strconv.ParseFloat`) and routes both `handleMemoryAdopt` and `handleMemoryConnect` through the unified `getFloat`. Good coverage improvement — confirmed via `internal/mcp/tools_typed_args_test.go`, which tests native float64, a JSON numeric string, and an uncoercible garbage string ("na", correctly falls back to default).

**Blocker (2) is not actually fixed.** `handleMemoryConnect` already had (pre-existing, untouched by this diff) a post-coercion guard:
```go
if math.IsNaN(strength) || math.IsInf(strength, 0) || strength < 0 || strength > 1.0 {
    return nil, fmt.Errorf("strength must be between 0 and 1, got %v", strength)
}
```
`handleMemoryAdopt` has NO equivalent guard, before or after this diff — it goes straight from `getFloat(args, "strength", 1.0)` to `h.Engine.Connect(...)`. Verified downstream: `SearchEngine.Connect` (`internal/search/engine.go:1835`) only checks `if strength <= 0 { strength = 1.0 }` — this does NOT catch NaN, because `NaN <= 0` is always `false` under IEEE 754. **A literal string `"NaN"` (or a numeric NaN, or +/-Inf) passed as `strength` to `memory_adopt` is parsed successfully by `strconv.ParseFloat` into a real NaN/Inf float and persisted directly into the relationship's `Strength` field with zero validation.** Confirmed by direct code read, not assumed. The new test file also does not cover this exact case — it tests `"na"` (fails to parse, falls back to default 1.0) but never the literal `"NaN"` string (which DOES parse successfully and is the actual reported bug).

## Fix needed
Add the same `math.IsNaN(strength) || math.IsInf(strength, 0) || strength < 0 || strength > 1.0` guard (returning an error) to `handleMemoryAdopt` in `internal/mcp/tools_admin.go`, mirroring `handleMemoryConnect`. Add a test case for the literal string `"NaN"` (and ideally `"Inf"`/`"-Inf"`) against BOTH `memory_adopt` and `memory_connect` in `tools_typed_args_test.go`, asserting an error is returned, not a silently-stored NaN/Inf.

petersimmons1972:
Codex work for #1307 via fleet-dispatch. Draft — needs review/CI. Closes #1307

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
+	"strconv"
 	"strings"
 	"time"
 	"unicode/utf8"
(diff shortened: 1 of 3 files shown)


petersimmons1972:
Closing — good logic (the validateRelationshipStrength approach and NaN/Inf-string tests are correct and were kept), but this branch is based on a stale main (same `ffa37bd` parent implicated in several other stale-base incidents tonight) and duplicates `coerceToFloat`/`getFloat`, which already landed on main via `073b1507` — hence the unresolvable rebase conflicts (content conflict in `tools.go`, add/add conflict on `tools_typed_args_test.go`).

Superseding fix dispatched as a minimal diff authored directly against current `main` (keeps `validateRelationshipStrength` + the `handleMemoryAdopt`/`handleMemoryConnect` wiring + the NaN/Inf-string tests in a new, non-conflicting test file; drops the duplicate coerce/getFloat redefinition) — will land as a NEW PR referencing #1307. Branch kept here for forensics, not deleted.

PR #1325, opened 2026-07-05
fix(longmemeval): sanitize control characters in LME-S ingest fixtures

### LINKED ISSUE #1321: defect: LME-S fixture contains raw control characters that memory_store_batch rejects
## Context

Found while running the real WP-0.5 retrofit fire-rate measurement (companion campaign, `petersimmons1972/duramind`, engram-go#1314/PR#1315) against a local isolated engram-go instance and the pinned 133-item LME-S multi-session fixture (`/tmp/lme_s_multisession_133.json`, gold `gold-lme-s-2026-07-03`).

## The defect

Question `e56a43b9`, haystack session 18, one `user`-role turn contains a raw U+0002 control character mid-word: `"...he major shared techniques including pre\x02dicting contextualized target represent..."`. This is upstream corpus noise (looks like a corrupted hyphenation/line-break artifact from a PDF-to-text extraction of an academic-paper distractor snippet embedded in the LME-S dataset), not something either harness introduces.

`memory_store_batch` correctly rejects it: `content: content contains disallowed control character U+0002 at byte 224`. This is arguably correct server-side input validation, not a bug in the handler itself.

## Why this matters for the campaign

`duramind`'s own greenfield ingest path (`internal/wp05brunner.IngestItem` -> `store.UpsertMemory`) has no equivalent content validation, so it silently ingested this same byte and the greenfield 133-item measurement completed with 0 skips. engram-go's retrofit runner (`internal/wp05retrofit`, this repo) hard-aborts the whole run on the first rejected batch item, which would otherwise make it impossible to get a real full-133 retrofit measurement without either (a) relaxing server-side validation (not desirable -- the validation itself looks correct) or (b) sanitizing benchmark-harness input before it reaches `Store`/`StoreBatch`.

## Recommended fix

Sanitize control characters (C0 range except `\t`/`\n`) out of rendered session content in `internal/wp05retrofit.renderSessionContent` before calling `Store`/`StoreBatch` -- this is a benchmark-harness-side fix (real user-facing content practically never contains raw C0 control bytes; this is LME corpus noise specific to the benchmark fixture), not a relaxation of the server's own input validation.

## What was done for THIS measurement

A throwaway, uncommitted, uncommitted-to-any-branch local patch to `renderSessionContent` (strip C0 control chars except tab/newline) was used to get the real fire-rate number without blocking on a full fleet-dispatch/review cycle for a one-line data-hygiene fix -- same pattern as `duramind#15`'s discovery (a throwaway DO-NOT-MERGE patch to keep a measurement moving while the real fix goes through proper review separately). Not merged, not pushed. This issue tracks the real fix as a small, low-risk follow-up whenever there's fleet-review capacity -- not campaign-critical-path.

petersimmons1972:
Fixes #1321: the LME-S fixture contained raw control characters that memory_store_batch rejects. Sanitizes control characters during ingest and adds regression coverage in cmd/longmemeval/ingest_test.go and internal/longmemeval/engram_test.go. (Title/body repaired by coordinator — the impl agent rebased the branch onto current main but left the placeholder title.)

### DIFF
--- a/cmd/longmemeval/ingest.go
+++ b/cmd/longmemeval/ingest.go
@@ -107,11 +107,17 @@ func ingestOne(ctx context.Context, cfg *Config, restClient *longmemeval.RestCli
 			continue
 		}
 		// Prepend session date so the model can anchor relative-time questions.
-		tags := []string{"lme", "sid:" + sessionID}
-		if i < len(item.HaystackDates) && item.HaystackDates[i] != "" {
-			content = "Session date: " + item.HaystackDates[i] + "\n" + content
-			tags = append(tags, "date:"+item.HaystackDates[i])
+		tags := []string{"lme"}
+		if cleanSessionID := longmemeval.SanitizeStoreContent(sessionID); cleanSessionID != "" {
+			tags = append(tags, "sid:"+cleanSessionID)
 		}
+		if i < len(item.HaystackDates) && item.HaystackDates[i] != "" {
+			if cleanDate := longmemeval.SanitizeStoreContent(item.HaystackDates[i]); cleanDate != "" {
+				content = "Session date: " + cleanDate + "\n" + content
+				tags = append(tags, "date:"+cleanDate)
+			}
+		}
+		content = longmemeval.SanitizeStoreContent(content)
 		sessions = append(sessions, sessionEntry{
 			sessionID: sessionID,
 			item:      longmemeval.BatchItem{Content: content, Tags: tags},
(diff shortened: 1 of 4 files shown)


petersimmons1972:
**fleet-dispatch** — item `89af1646-8c4f-47ed-9088-8490c10269e8` reached state `done`.

Result reference: `VERDICT: pass BLOCKERS: none NICE-TO-HAVES: 1. Add a test for C1 controls, e.g. U+0080/U+009F encoded in fixture text, since the sanitizer explicitly claims to remove both C0 and C1 ranges but the new ingest test only exercises a C0 byte. 2. Add a test where session ID or date sanitizes to empty, to lock in the intended behavior of omitting the corresponding tag/date prefix rather than storing 'sid:' or 'date:'.`

PR #1352, opened 2026-07-06
fix: make instinct healthcheck timestamp parsing portable

### LINKED ISSUE #1335: [adversarial audit] ops healthcheck uses GNU-only grep -P despite portability fallback
## Finding
`ops/healthcheck.sh` uses GNU-only `grep -P` for timestamp extraction even though it contains BSD/macOS date fallback logic.

## Evidence
- Persona(s): Operator/SRE, Heavy CLI User
- Commit audited: `7cf7433360465f255b4085e67dabb54ef3fd64fe`
- `ops/healthcheck.sh:18` uses `grep -oP`.
- `ops/healthcheck.sh:25` includes GNU `date -d` and BSD/macOS `date -j -f` fallback.

## Observed risk / failure mode
On macOS/BSD userlands, grep fails before the BSD date fallback can help, so the healthcheck incorrectly reports stale/missing state.

## Suggested Fix
Replace `grep -oP` with portable sed/awk or Python timestamp extraction.

## Acceptance Criteria
- [ ] Timestamp extraction works on GNU and BSD/macOS environments.
- [ ] Fixture log test runs without GNU grep.

## Verification
```bash
bash -n ops/healthcheck.sh
printf 'instinct: ok [2026-04-22T10:00:00Z]
' > /tmp/instinct-run.log
```

FM-16 note: no Docker/Kubernetes verification was run during this audit; suggested Docker/K8s commands, if later used, must clean up only resources they create.

petersimmons1972:
## Summary
- replaces GNU-only `grep -P` timestamp extraction in `ops/healthcheck.sh` with portable `sed`
- adds a shell regression test with fake grep rejecting `-P`
- removes the script's final unconditional success `exit 0` while preserving successful fallthrough

Closes #1335.

## Verification
- `bash ops/healthcheck.test.sh`
- `bash -n ops/healthcheck.sh ops/healthcheck.test.sh`
- `git diff --check`
- pre-commit `checkin-lint` passed during commit

### DIFF
--- a/ops/healthcheck.sh
+++ b/ops/healthcheck.sh
@@ -15,7 +15,7 @@ fi
 
 # Extract the most recent ISO-8601 timestamp from log lines of the form:
 #   instinct: ... [2026-04-22T10:00:00Z]
-last_ts=$(grep -oP '\[\K\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z(?=\])' "$LOG" 2>/dev/null | tail -1 || :)
+last_ts=$(sed -n 's/.*\[\([0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]T[0-9][0-9]:[0-9][0-9]:[0-9][0-9]Z\)\].*/\1/p' "$LOG" | tail -1 || :)
 
 if [[ -z "$last_ts" ]]; then
     echo "WARN: no timestamped instinct runs found in $LOG"
@@ -34,4 +34,3 @@ fi
 
 echo "OK: instinct last ran ${gap}s ago"
 echo "Last: $last_ts"
-exit 0
--- a/ops/healthcheck.test.sh
+++ b/ops/healthcheck.test.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+# Test suite for ops/healthcheck.sh
+# Usage: bash ops/healthcheck.test.sh
+
+set -u
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SCRIPT="$SCRIPT_DIR/healthcheck.sh"
+
+PASS=0
+FAIL=0
+
+assert_exit() {
+  local desc="$1" expected="$2" actual="$3"
+  if [ "$expected" -eq "$actual" ]; then
+    echo "PASS: $desc (exit=$actual)"
+    PASS=$((PASS + 1))
+  else
+    echo "FAIL: $desc (expected=$expected got=$actual)"
+    FAIL=$((FAIL + 1))
+  fi
+}
+
+assert_contains() {
+  local desc="$1" needle="$2" haystack="$3"
+  if echo "$haystack" | grep -qF -- "$needle"; then
+    echo "PASS: $desc"
+    PASS=$((PASS + 1))
+  else
+    echo "FAIL: $desc (missing '$needle')"
+    FAIL=$((FAIL + 1))
+  fi
+}
+
+tmp="$(mktemp -d)"
+trap 'rm -rf "$tmp"' EXIT
+mkdir -p "$tmp/bin" "$tmp/state/instinct"
+
+cat >"$tmp/bin/grep" <<'FAKE_GREP'
+#!/usr/bin/env bash
+for arg in "$@"; do
+  case "$arg" in
+    *P*) echo "fake grep: -P is unavailable" >&2; exit 2 ;;
+  esac
+done
+exec /usr/bin/grep "$@"
+FAKE_GREP
+chmod +x "$tmp/bin/grep"
+
+now_ts="$(date -u +"%Y-%m-%dT%H:%M:%SZ")"
+printf 'instinct: ok [%s]\n' "$now_ts" >"$tmp/state/instinct/run.log"
+
+out=$(PATH="$tmp/bin:$PATH" XDG_STATE_HOME="$tmp/state" INSTINCT_MAX_GAP=7200 bash "$SCRIPT" 2>&1)
+rc=$?
+assert_exit "healthcheck works without grep -P" 0 "$rc"
+assert_contains "reports healthy status" "OK: instinct last ran" "$out"
+assert_contains "reports extracted timestamp" "Last: $now_ts" "$out"
+
+echo
+echo "--- ${PASS} passed, ${FAIL} failed ---"
+[ "$FAIL" -eq 0 ]


petersimmons1972:
**Fleet adversarial review (fleet-dispatch item `d12b8618-8097-4b95-85fd-771a7d6e744b`, capability=review, diff-inline protocol)**

**VERDICT: fail**

**BLOCKERS:**
1. When `$LOG` does not exist or is unreadable (fresh container/volume, log rotation, wrong XDG_STATE_HOME, or instinct has never written yet), the new `sed ... "$LOG"` emits "sed: can't read ...: No such file or directory" on stderr (original had `2>/dev/null`). `last_ts` is empty, the WARN line prints, and (per the patch hunks) control falls through to the gap-calculation code that follows the `if [[ -z "$last_ts" ]]` block. This produces `date: invalid date ''`, arithmetic errors in `$((...))`, or garbage `${gap}`, so the healthcheck either emits confusing output or exits non-zero (or falsely prints an "OK" line). The test never creates this scenario.
2. The test writes exactly one log line and asserts only the happy path. The entire "most recent" contract depends on `... | tail -1` over extracted timestamps from the whole file. No test feeds multiple chronologically ordered lines and asserts that the newest timestamp (and only that one) appears in the "Last:" / OK output. A sed change that emitted only the first match, or that only processed early lines, would still pass the test while reporting a stale run as healthy.
3. The final `exit 0` was deleted. The OK path now ends with two `echo` commands. If the last `echo "Last: $last_ts"` fails (SIGPIPE to closed stdout, wrapper that exits non-zero on write error, etc.), the script exits non-zero even though the logic decided the service was healthy. Explicit `exit 0` made the healthy contract independent of the last command's side effects.

**NICE-TO-HAVES:**
1. Add at least one multi-line test case (older entry first) that asserts the newest timestamp is the one reported.
2. Add a test case that exercises the empty-`last_ts` / WARN path and asserts the resulting exit code (presumably non-zero) plus absence of date/arithmetic garbage.
3. Add a test case with `INSTINCT_MAX_GAP` set low + an old timestamp that asserts the FAIL / non-zero path.
4. Either restore `exit 0` on the success path or add `2>/dev/null` (or an explicit existence check) to the sed invocation to preserve the prior quiet behavior on missing logs.
5. Make the fake-grep shim stricter or add an explicit `! command -v grep -oP ...` style assertion, and add a test line containing two bracketed timestamps to lock in "last on line" behavior.
6. Consider `tail -n 1` (slightly more modern/portable spelling) and a brief comment explaining why BRE sed was chosen.

petersimmons1972:
**fleet-dispatch** — item `43b5e0e0-2112-4cd0-89d2-1dd4e6589b27` reached state `done`.

Result reference: `APPROVE  No concrete blockers found within the stated non-LME scope.  Review focus:  - Shell portability:   - Replacing 'grep -oP '\[\K...(?=\])'' with POSIX-style 'sed' capture avoids the GNU-only 'grep -P' dependency that caused BSD/macOS failure.   - The sed approach described is compatible with BSD/macOS sed syntax: 'sed -n 's/.*\[\(...\)\].*/\1/p''.   - 'tail -1' works on GNU and BSD/macOS in practice. If the project wants stricter POSIX style, 'tail -n 1' would be cleaner, but I would not block on it for the stated BSD/macOS target.  - Timestamp extraction correctness:   - Extracting bracketed ISO/Zulu timestamps from log lines and taking the last match is consistent with the intended healthcheck freshness check.   - The pattern is appropriately constrained to timestamps ending in 'Z', so it should not accidentally extract arbitrary bracketed log metadata unless it also matches the timestamp shape.   - Greedy '.*\[' means it selects the last matching bracketed timestamp on a line. That is acceptable for…`

petersimmons1972:
**fleet-dispatch** — item `1199d6e2-b0b7-4aa8-b826-5c5ce7d0522f` reached state `done`.

Result reference: `**APPROVE**  No blockers. The PR correctly resolves #1335 for the stated scope (only 'ops/healthcheck.sh' + new 'ops/healthcheck.test.sh').  ### Shell portability - 'grep -oP '...' 2>/dev/null | tail -1 || :' (PCRE, GNU grep only) → 'sed -n 's/.*\[\([0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]T[0-9][0-9]:[0-9][0-9]:[0-9][0-9]Z\)\].*/\1/p' "$LOG" | tail -1 || :' - BRE 'sed -n s//p' + 'tail -1' (and the rest of the script) is portable across GNU + BSD/macOS userlands. The existing 'date -d' / 'date -j -f' fallback is now reachable. - Script remains '#!/usr/bin/env bash' + 'set -euo pipefail' + '[[' / '((' (appropriate; not claiming POSIX sh). - 'tail -1' (vs. '-n 1') is unchanged from the prior version and de-facto portable.  ### Timestamp extraction correctness - The sed correctly extracts the *last* matching '[YYYY-MM-DDTHH:MM:SSZ]' per the comment/example (greedy '.*\[...\].*' + '| tail -1' for chronological last line). - Verified:   - Single match, multiple matches per line (picks rightmost/last), no-match → …`

petersimmons1972:
**fleet-dispatch** — item `189731f8-35c7-4a26-83b9-fa13f220093c` reached state `done`.

Result reference: `**APPROVE**  No concrete blockers. The PR correctly solves #1335 with portable code and a test that actually enforces the fix.  ### Summary of changes (from commit 9be16ad) - 'ops/healthcheck.sh': replace 'grep -oP '\[\K\d{4}-...Z(?=\])' ...' with:   '''sh   sed -n 's/.*\[\([0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]T[0-9][0-9]:[0-9][0-9]:[0-9][0-9]Z\)\].*/\1/p' "$LOG" | tail -1 || :)   ''' - Remove the final unconditional 'exit 0' (fallthrough on the last 'echo'). - New 'ops/healthcheck.test.sh': fake 'grep' that rejects any arg containing 'P', fresh 'date -u' fixture in a temp log, run healthcheck under 'PATH' override, assert exit 0 + healthy output + exact timestamp echoed.  ### Verification performed - 'bash ops/healthcheck.test.sh' (using exact committed versions) → 3/3 PASS, exit 0. - 'bash -n' on both files → clean. - 'git diff --check' → clean. - Reproduced the core claim: when the extraction is forcibly reverted to 'grep -oP', the same test setup yields rc=1 ("no timestamped...") because the fake gre…`

petersimmons1972:
**fleet-dispatch** — item `d12b8618-8097-4b95-85fd-771a7d6e744b` reached state `done`.

Result reference: `**VERDICT: fail**  **BLOCKERS:** 1. When '$LOG' does not exist or is unreadable (fresh container/volume, log rotation, wrong XDG_STATE_HOME, or instinct has never written yet), the new 'sed ... "$LOG"' emits "sed: can't read ...: No such file or directory" on stderr (original had '2>/dev/null'). 'last_ts' is empty, the WARN line prints, and (per the patch hunks) control falls through to the gap-calculation code that follows the 'if [[ -z "$last_ts" ]]' block. This produces 'date: invalid date ''', arithmetic errors in '$((...))', or garbage '${gap}', so the healthcheck either emits confusing output or exits non-zero (or falsely prints an "OK" line). The test never creates this scenario. 2. The test writes exactly one log line and asserts only the happy path. The entire "most recent" contract depends on '... | tail -1' over extracted timestamps from the whole file. No test feeds multiple chronologically ordered lines and asserts that the newest timestamp (and only that one) appears in the "Last:" / OK output. …`

PR #1363, opened 2026-07-07
feat(longmemeval): add full timeline benchmark mode

petersimmons1972:
## Summary - add benchmark-only `--full-timeline-context` support to `longmemeval run`, bypassing `memory_recall`/`memory_fetch` and injecting the full dated haystack timeline into generation prompts - persist `generation_context` in `run_manifest.json` and `RUN_STATUS.json`, preserving prior run-stage attribution during later score-only invocations - document the new mode and lock it with focused CLI, orchestration, and prompt-path tests ## Testing - `/usr/local/go/bin/go test ./cmd/longmemeval` - `/usr/local/go/bin/go test ./...` *(sandbox failure: Postgres `vector` extension unavailable; pre-existing failures in `internal/consolidate`, `internal/db`, and `internal/search`)* - `./bin/checkin-lint.sh`


petersimmons1972:
**fleet-dispatch** — item `3ec2d135-a65c-4378-a144-253836275fee` reached state `done`.

Result reference: `VERDICT: fail BLOCKERS: 1. Full-timeline mode can silently stop being a “timeline” for questions where 'targetDateFromQuestion(...)' succeeds. In 'runOne', the ordering branch applies 'sortBlocksByTargetDate(...)' before the new 'fullTimelineContext' chronological branch:    'if targetDateFromQuestion(...) { ... } else if fullTimelineContext { sortBlocksChronologically(...) }'.    That means full-timeline context may be reordered around the target date instead of preserving chronological haystack order. Failure scenario: a temporal/date-bearing question with sessions before and after the target date gets prompt context ordered by target-date proximity rather than as the full dated timeline, changing the benchmark condition and making A/B attribution misleading. Full-timeline mode should probably bypass target-date proximity sorting and always use dataset order or chronological date order.  2. 'fullTimelineContextBlocks' truncates each session after adding the 'Session date:' header. With '--max-block-chars' s…`

PR #1364, opened 2026-07-07
fix(ops): point instinct.service at installed Go consolidator

petersimmons1972:
## Summary\n- update `ops/instinct.service` to run `%h/bin/instinct-consolidate`, matching `make install-instinct`\n- remove the stale `PYTHONPATH`/`python3 -m instinct.run` wiring from the timer-mode unit\n- add `ops/instinct-service.test.sh` as a focused regression test for the service contract\n\n## Testing\n- `bash ops/instinct-service.test.sh` ✅\n- `bash -n ops/instinct-service.test.sh` ✅\n- `go test ./cmd/instinct ./cmd/audit -count=1` ✅\n- temporary HOME install smoke check via `make install-instinct` with disposable `HOME/bin` ✅\n- `git diff --check` ✅\n- `bash bin/checkin-lint.sh` ✅\n\n## Environment notes\n- `systemd-analyze verify ops/instinct.service` could not run in the sandbox because `systemd-analyze` is not installed\n- the exact `GOMODCACHE=/home/psimmons/go/pkg/mod` install check was not writable in this sandbox, so the install smoke check used the available module cache at `/opt/gomodcache`\n- the broad non-LME package sweep hit pre-existing environment failures unrelated to this change (`pgvector` extension unavailable and an existing `internal/embed` concurrency test failure)

### DIFF
--- a/bin/checkin-lint.baseline
+++ b/bin/checkin-lint.baseline
@@ -246,7 +246,7 @@ D.exit-zero-wrapper::./.worktrees/issue-1063-extract-parser/hooks/engram/engram-
 D.exit-zero-wrapper::./.worktrees/issue-1063-extract-parser/hooks/engram/engram-precheck.sh::38
 D.exit-zero-wrapper::./.worktrees/issue-1063-extract-parser/hooks/pre-tool-use.sh::4
 D.exit-zero-wrapper::./ops/healthcheck.sh::37
-D.exit-zero-wrapper::./hooks/post-tool-use.sh::111
+D.exit-zero-wrapper::./hooks/post-tool-use.sh::113
 D.exit-zero-wrapper::./hooks/engram/engram-session-end.sh::16
 D.exit-zero-wrapper::./hooks/engram/engram-session-end.sh::65
 D.exit-zero-wrapper::./hooks/engram/engram-token-refresh.sh::33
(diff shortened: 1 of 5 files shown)


petersimmons1972:
**fleet-dispatch** — item `e1be915c-6497-4061-8b33-3c4cb2798fbf` reached state `done`.

Result reference: `**VERDICT: fail**  **BLOCKERS:** 1. Binary name mismatch is not resolved. The full diff touches only 'ops/instinct.service' and the new test. 'hooks/install.sh' still does 'go build -o "$HOME/bin/instinct"' and 'hooks/post-tool-use.sh' still does 'command -v instinct'. Service + test now hard-require '%h/bin/instinct-consolidate'. Any user who installs via the documented hook path ends up with a permanently failing oneshot ('No such file or directory' on timer tick, triggering 'engram-instinct-alert.service'). 2. Setup instructions remain incomplete and now actively point to a non-existent step. The diff adds only the single comment line '#   make install-instinct'. No Makefile diff, no 'go build -o %h/bin/instinct-consolidate' anywhere, and 'ops/README.md' is untouched. A user who follows the header comments exactly runs a make target that does not exist in this PR (or gets the wrong binary name), so the service never starts. 3. The contract test is still only a source grep. 'ops/instinct-service.test.sh' do…`

PR #1357, opened 2026-07-06
fix: fail loudly on load-bearing optional params

petersimmons1972:
## Summary

Closes #1282.

This is a bounded first pass on load-bearing optional parameter validation. It changes parameters that can materially affect migration, audit scope, or sleep/consolidation cost from silent-default behavior to loud typed validation errors.

Covered params:
- `memory_migrate_embedder`: `force`, `dry_run`, `confirm`
- `memory_audit_compare`: `limit`
- `memory_sleep`: `limit`, `llm_max_calls`, `contradiction_limit`, `llm_contradiction_detection`, `auto_supersede`

Out of scope for this PR:
- Retrieval/tuning-only params that are not load-bearing for this issue.
- LME/WP05-adjacent work under the current stop boundary.
- Runtime/container changes.

## Verification

Red tests before implementation:
- `requireOptionalBool` helper test failed before the helper existed.
- `memory_migrate_embedder` wrong-type guard tests failed because bool args silently defaulted.
- `memory_audit_compare limit` tests failed because string/wrong-type input was not validated as intended.
- `memory_sleep` wrong-type tests reached pool access instead of returning validation errors.

Green tests after implementation:
- `go test ./internal/mcp -run 'TestRequireOptionalBool|TestMCPMigrate.*WrongTypeReturnsLoudError|TestMCPMigrate.*ArgPassedToEngine|TestHandleAuditCompare_Limit|TestMemorySleepLoadBearingParamsWrongTypeReturnLoudError' -count=1`
- `go test ./internal/mcp -count=1`
- `pkgs=$(go list ./... | grep -Ev '/(cmd/longmemeval|internal/longmemeval|internal/wp05retrofit|cmd/wp05-retrofit-runner)(/|$)') && go test $pkgs -count=1`
- `git diff --check`
- checkin-lint via commit hook

## Manifest First

No container, pod, service unit, manifest, or runtime behavior was changed by this PR.

PR #1365, opened 2026-07-07
fix(mcp): validate load-bearing optional params

petersimmons1972:
Codex work for #1357 via fleet-dispatch. Draft — needs review/CI. Closes #1357

### DIFF
--- a/internal/mcp/migrate_guards_test.go
+++ b/internal/mcp/migrate_guards_test.go
@@ -18,6 +18,16 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func requireToolErrorText(t *testing.T, result *mcpgo.CallToolResult) string {
+	t.Helper()
+	require.NotNil(t, result)
+	require.True(t, result.IsError, "expected MCP tool error result")
+	require.NotEmpty(t, result.Content)
+	text, ok := result.Content[0].(mcpgo.TextContent)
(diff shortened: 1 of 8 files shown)


petersimmons1972:
**fleet-dispatch** — item `077055d3-47e9-45a7-ae15-5e08df9ec1a1` reached state `done`.

Result reference: `**VERDICT: fail**  **BLOCKERS:** 1. 'handleMemorySleep' still silently accepts a wrong-typed 'min_similarity' via 'getFloat(args, "min_similarity", 0.7)'. Concrete failure scenario: a client sends '"min_similarity": []' or another non-float value intending a stricter threshold; the handler falls back to '0.7' and proceeds (see the unchanged 'minSim := getFloat' line + RunAll call in tools_consolidate.go), potentially consolidating/superseding a much broader set of memories than intended instead of failing loudly. 2. 'handleMemorySleep' still silently accepts a wrong-typed 'llm_model' via 'getString(args, "llm_model", "llama3.2:3b")'. Concrete failure scenario: with '"llm_contradiction_detection": true' (now correctly extracted via requireOptionalBool), a malformed '"llm_model": []' request silently falls back to 'llama3.2:3b' (see the unchanged 'llmModel := getString' line), causing contradiction detection to run against the wrong model rather than rejecting the request. 3. The new 'TestMemorySleepLoadBearing…`

PR #1358, opened 2026-07-06
fix: label dropped recall hits accurately

petersimmons1972:
## Summary

Closes #1312.

This PR separates recall degradation warning text for two distinct paths:
- Embed/health degradation still emits `recall degraded: embed unavailable, using BM25 fallback`.
- Nil-Memory backend hits now emit `recall degraded: dropped backend hits with missing memory records` instead of implying an embed fallback.

It also adds the missing regression test for temporal two-pass dropped-hit accumulation: one nil-Memory FTS result in each pass must produce `DroppedHits == 2`.

## Verification

Red tests before implementation:
- `go test ./internal/mcp -run TestHandleMemoryRecall_NilMemoryResultSurfacesDroppedHitsInDegradedField -count=1` failed before the dropped-hit warning existed.
- The temporal two-pass test required the new expectation that dropped hits accumulate from both passes.

Green tests after implementation:
- `go test ./internal/mcp -run 'TestHandleMemoryRecall_NilMemoryResultSurfacesDroppedHitsInDegradedField|TestHandleMemoryRecall_.*Embed.*Degraded' -count=1`
- `go test ./internal/search -run TestRecallTwoPassTemporal_AccumulatesDroppedHitsFromBothPasses -count=1`
- `go test ./internal/mcp ./internal/search -count=1`
- `pkgs=$(go list ./... | grep -Ev '/(cmd/longmemeval|internal/longmemeval|internal/wp05retrofit|cmd/wp05-retrofit-runner)(/|$)') && go test $pkgs -count=1`
- `git diff --check`
- checkin-lint via commit hook

## Manifest First

No container, pod, service unit, manifest, or runtime deployment behavior was changed by this PR.

PR #1366, opened 2026-07-07
fix(mcp): split dropped-hit recall warnings

petersimmons1972:
Codex work for #1358 via fleet-dispatch. Draft — needs review/CI. Closes #1358


petersimmons1972:
**fleet-dispatch** — item `838c7e39-3b31-4c50-a2a3-b934f9e6123b` reached state `done`.

Result reference: `VERDICT: fail  BLOCKERS: 1. Original blocker 3 is still not resolved. The MCP handler coverage still does not assert warning list length, order, or uniqueness, and it still does not exercise the full handler path where embed degradation and dropped hits are both present. The added fetch test only checks Contains/NotContains for dropped-only, so it would still miss duplicated dropped warnings or other extra warnings. The new temporal engine test only verifies DroppedHits accumulation below the MCP layer and never exercises addRecallWarnings or the final MCP response shape.  2. The duplicate-call shape from original blocker 1 is only partially addressed. The final warnings field is less likely to contain duplicate strings because addRecallWarnings overwrites out["warnings"] with a fresh slice, but the handler still has two addRecallWarnings call sites in the same invocation path. In an embed-degraded recall that reaches both sites, the same degraded warning/log path can still be emitted twice via slog.Warn. Tha…`

